PR #4, opened 2016-11-17
Add Parse func accepting various types

lgierth:
@Kubuxu @whyrusleeping @keks for review

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"strings"
 
 	mbase "github.com/multiformats/go-multibase"
 	mh "github.com/multiformats/go-multihash"
@@ -45,6 +46,24 @@ type Cid struct {
 	version uint64
 	codec   uint64
 	hash    mh.Multihash
+}
+
+func Parse(v interface{}) (*Cid, error) {
+	switch v2 := v.(type) {
+	case string:
+		if strings.Contains(v2, "/ipfs/") {
+			return Decode(strings.Split(v2, "/ipfs/")[1])
+		}
+		return Decode(v2)
+	case []byte:
+		return Cast(v2)
+	case mh.Multihash:
+		return NewCidV0(v2), nil
+	case *Cid:
+		return v2, nil
+	default:
+		return nil, fmt.Errorf("can't parse %+v as Cid", v2)
+	}
 }
 
 func Decode(v string) (*Cid, error) {
--- a/cid_test.go
+++ b/cid_test.go
@@ -2,7 +2,9 @@ package cid
 
 import (
 	"bytes"
+	"fmt"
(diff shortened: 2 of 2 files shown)


lgierth:
For explanation, this func is useful in Core API, where we want to be flexible in what the we expect as function arguments. Right now we achieve this flexibility through requiring strings, but that's less than ideal as it involved lots of casting from-and-to. With this change, the Core API functions can just accept `interface{}`, and let the cid package take care of turning that into a Cid object.

cid.go
```diff
@@ -47,6 +48,25 @@ type Cid struct {
 	hash    mh.Multihash
 }
 
+func Parse(v interface{}) (*Cid, error) {
+	switch v.(type) {
+	case string:
+		v2 := v.(string)
+		if strings.Contains(v2, "/ipfs/") {
+			return Decode(strings.Split(v2, "/ipfs/")[1])
```
lgierth: Note that this also accepts https://ipfs.io/ipfs/hash :)

keks:
You *could* check the errors returned by `Parse`. If tests pass there is no difference, but if they don't the error value gives you guidance what the problem is.

Else LGTM.

cid.go
```diff
@@ -47,6 +48,25 @@ type Cid struct {
...
+		}
+		return Decode(v2)
+	case []byte:
+		return Cast(v.([]byte))
+	case mh.Multihash:
+		return NewCidV0(v.(mh.Multihash)), nil
+	case *Cid:
+		return v.(*Cid), nil
+	default:
+		return nil, fmt.Errorf("can't parse %+v as Cid", v)
```
Kubuxu: but it might be reference to struct name

lgierth: Yep that's referencing the struct we're trying to get out of this

lgierth:
> You could check the errors returned by Parse. If tests pass there is no difference, but if they don't the error value gives you guidance what the problem is.

Yep updated, good point

cid.go
```diff
@@ -47,6 +48,25 @@ type Cid struct {
 	hash    mh.Multihash
 }
 
+func Parse(v interface{}) (*Cid, error) {
+	switch v.(type) {
```
lgierth: > Instead of casting everywhere you can do switch with assignment, switch value := v.(type) {.

Updated

cid.go
```diff
@@ -47,6 +48,25 @@ type Cid struct {
 	hash    mh.Multihash
 }
 
+func Parse(v interface{}) (*Cid, error) {
```
lgierth: > it can give imprecise result

I don't follow, have I overlooked something?

lgierth:
Also DRY'ed up the tests

cid.go
Kubuxu: Nah, doesn't matter.

PR #6, opened 2016-11-17
Expose panic in Cast()

lgierth:
Not sure exactly how to fix it yet, but I though I'd write it down for general entertainment.

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -88,6 +88,7 @@ func Cast(data []byte) (*Cid, error) {
 		return nil, fmt.Errorf("invalid cid version number: %d", vers)
 	}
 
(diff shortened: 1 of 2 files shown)


whyrusleeping:
@lgierth the only way i see this happening is if theres a race condition. How are you reproducing the panic?

binary.Uvarint won't return a value past the length of the buffer.

lgierth:
`while true; do go test . ; done` crashes 100 % of the runs

lgierth:
It starts crashing for some `500 < i < 1000`

lgierth:
Taken care of in #7

PR #7, opened 2016-11-17
Add fuzz tests, fix panic

Kubuxu:


Kubuxu:
There are two more issues with JSON encoding.
There is a case when if fails to encode data and there is case when it fails to decode data which it has encoded.

Kubuxu:
```
==> 576e1bc17d21d855785ef74b9a35ab9cf7c04fe0 <==
0 00000000000000000000000000000000
==> 576e1bc17d21d855785ef74b9a35ab9cf7c04fe0.output <==
panic: json loop not equal

goroutine 1 [running]:
panic(0x4d92c0, 0xc42000a420)
	/tmp/go-fuzz-build301216780/src/runtime/panic.go:500 +0x1a1
github.com/ipfs/go-cid.Fuzz(0x7ff7ff6f2000, 0x24, 0x200000, 0x0)
	/tmp/go-fuzz-build301216780/src/github.com/ipfs/go-cid/cid_fuzz.go:32 +0x32f
go-fuzz-dep.Main(0x505da0)
	/tmp/go-fuzz-build301216780/src/go-fuzz-dep/main.go:49 +0xe8
main.main()
	/tmp/go-fuzz-build301216780/src/go.fuzz.main/main.go:10 +0x2d
exit status 2
==> 576e1bc17d21d855785ef74b9a35ab9cf7c04fe0.quoted <==
	"\x000\x12 0000000000000000" +
	"0000000000000000"

==> e79d011f70c39fb73dbecd6fc4a66684d5a289b5 <==
0 00000000000000000000000000000000
==> e79d011f70c39fb73dbecd6fc4a66684d5a289b5.output <==
panic: selected encoding not supported

goroutine 1 [running]:
panic(0x4d92c0, 0xc42000a420)
	/tmp/go-fuzz-build301216780/src/runtime/panic.go:500 +0x1a1
github.com/ipfs/go-cid.Fuzz(0x7febf9c0c000, 0x24, 0x200000, 0x0)
	/tmp/go-fuzz-build301216780/src/github.com/ipfs/go-cid/cid_fuzz.go:28 +0x3a2
go-fuzz-dep.Main(0x505da0)
	/tmp/go-fuzz-build301216780/src/go-fuzz-dep/main.go:49 +0xe8
main.main()
	/tmp/go-fuzz-build301216780/src/go.fuzz.main/main.go:10 +0x2d
exit status 2
==> e79d011f70c39fb73dbecd6fc4a66684d5a289b5.quoted <==
	"\x000\x03 0000000000000000" +
	"0000000000000000"
```

cid.go
```diff
@@ -187,7 +187,8 @@ func (c *Cid) bytesV0() []byte {
 }
 
 func (c *Cid) bytesV1() []byte {
-	buf := make([]byte, 8+len(c.hash))
+	// two 8 bytes (max) numbers plus hash
+	buf := make([]byte, 2*8+len(c.hash))
```
Kubuxu: I was looking for it, didn't find it. Thanks.

cid.go
```diff
@@ -162,7 +187,8 @@ func (c *Cid) bytesV0() []byte {
 }
 
 func (c *Cid) bytesV1() []byte {
-	buf := make([]byte, 8+len(c.hash))
+	// two 8 bytes (max) numbers plus hash
+	buf := make([]byte, 2*binary.MaxVarintLen64+len(c.hash))
 	n := binary.PutUvarint(buf, c.version)
 	n += binary.PutUvarint(buf[n:], c.codec)
 	copy(buf[n:], c.hash)
```
whyrusleeping: We should take and check the return value of this copy to be extra safe. Its fine to panic if its different than we expect, but we don't want to accidentally return less than the whole thing

PR #9, opened 2016-11-20
fix raw codec number to be 0x55

whyrusleeping:
cc @Kubuxu @diasdavid to double check the rest

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -14,10 +14,10 @@ import (
 const UnsupportedVersionString = "<unsupported cid version>"
 
 const (
-	Protobuf = 0x70
-	CBOR     = 0x71
-	Raw      = 0x72
-	JSON     = 0x73
+	Raw = 0x55
+
+	DagProtobuf = 0x70
+	DagCBOR     = 0x71
 
 	EthereumBlock = 0x90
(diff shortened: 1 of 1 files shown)


Kubuxu:
There is no dag-json format in multicodec table.

cid.go
```diff
@@ -14,9 +14,10 @@ import (
 const UnsupportedVersionString = "<unsupported cid version>"
 
 const (
+	Raw = 0x55
```
diasdavid: 👍

cid.go
```diff
@@ -14,9 +14,10 @@ import (
...
+
 	Protobuf = 0x70
 	CBOR     = 0x71
-	Raw      = 0x72
 	JSON     = 0x73
```
whyrusleeping: Ehh... probably not.

kevina:
Um, this is a breaking change.  After this change when I tried to use raw leaves I now get unrecognized object type.

Is there any way to fix an repo with the incorrect numbers?

whyrusleeping:
It's a breaking change only between two unreleased versions. You can do a
file renaming in the blocks directory to fix it.

PR #12, opened 2016-11-23
fix build failures i introduced

whyrusleeping:
So it turns out we don't have CI enabled here...


lgierth:
I dunno, it looks like we still don't have CI enabled

lgierth:
> LGTM, but enable CI before merge. :smile:

:+1:

Kubuxu:
but let's wait for CI.

Issue #14, opened 2016-12-19
Serializable Cid type - Exporting .Version .Codec .Hash


hsanjuan:
The Cid-Type fields are unexported:

https://github.com/ipfs/go-cid/blob/master/cid.go#L46

This means that Cids need to be stringified/jsonified etc. when serialized (encoded/decoded) and re-decoded when deserialized.

What do people think of exporting those fields?

hsanjuan:
See #16

PR #16, opened 2017-01-20
Issue #14: Make Cids serializable by exporting their fields.

hsanjuan:
This is very useful when sending Cids as part of RPCs.

License: MIT
Signed-off-by: Hector Sanjuan <hector@protocol.ai>

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -29,24 +29,24 @@ const (
 
 func NewCidV0(h mh.Multihash) *Cid {
 	return &Cid{
-		version: 0,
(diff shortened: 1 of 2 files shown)


jbenet:
CIDs have a specific way to be serialized. go-cid already has this: https://github.com/ipfs/go-cid/blob/master/cid.go#L174

CIDs should not be serialized / passed on the wire in other ways (eg automatic json or msgpack serialization). It should always be the properly encoded CID token.

hsanjuan:
hm ok

PR #26, opened 2017-06-30
Added map of string to codecs

hermanjunge:
Example

```go
cid.Codecs["eth-block"] // Will evaluate to '\x90' (cid.EthBlock)
```

Very useful for `ipfs block put --format <format> --mhtype <mhtype> <object>`

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -80,7 +80,7 @@ const (
 )
 
 // Codecs maps the name of a codec to its type
-var Codecs = map[string]byte{
+var Codecs = map[string]uint64{
 	"v0":                   DagProtobuf,
 	"raw":                  Raw,
 	"protobuf":             DagProtobuf,
--- a/cid.go
+++ b/cid.go
@@ -79,6 +79,28 @@ const (
 	ZcashTx            = 0xc1
 )
 
+// Codecs maps the name of a codec to its type
+var Codecs = map[string]byte{
+	"v0":                   DagProtobuf,
+	"raw":                  Raw,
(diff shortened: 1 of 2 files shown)


cid_test.go
```diff
@@ -26,6 +49,22 @@ func assertEqual(t *testing.T, a, b *Cid) {
 	}
 }
 
+func TestTable(t *testing.T) {
+	for k, v := range tCodes {
+		if Codecs[v] != k {
+			t.Errorf("Table mismatch: 0x%x %s", k, v)
+		}
```
Stebalien: There should probably be a `len(tCodes) == len(Codecs) - 1` test here.

hermanjunge: @Stebalien improved the tests. Added the one requested.

cid.go
```diff
@@ -79,6 +79,28 @@ const (
 	ZcashTx            = 0xc1
 )
 
+// Codecs maps the name of a codec to its type
+var Codecs = map[string]byte{
```
Kubuxu: I am not sure that byte is right format for it.
The codecs are not limited to a byte (due to varint) we just are trying to keep codecs that are being used in cid in the 8 bit range but it might not be much longer till we will have two byte CID codecs (we already have two and three byte multihashes).

Kubuxu: It should probably be `uint64` as in https://github.com/ipfs/go-cid/blob/master/cid.go#L109

hermanjunge: @Kubuxu fixed

PR #29, opened 2017-07-08
gx publish 0.7.17

Stebalien:
Update go-multihash (to include the ID hash format).


whyrusleeping:
ooooh... we could start doing 'identity' blocks now. iiiiinteresting

Stebalien:
Yep. I look forward to seeing how people abuse it :smile:; one can use it to store an entire IPLD DAG in a single block! Personally, I'm using it to generate valid CIDs for test cases without associated blocks but that's a bit boring...

This also opens up a nice world of prefix-free typed values (e.g., `CIDv1-cbor-id-length-[cbor type][cbor data]`).

PR #31, opened 2017-08-15
Implement 'cid-fmt' utility.

kevina:
Useful in debugging.

This utility can also be used to change the base of a CID and CID version.

It will be very useful in go-ipfs sharness tests so we can check that the right Cid prefix is used without having to hardcode hashes.

### DIFF
--- a/cid-fmt/main.go
+++ b/cid-fmt/main.go
@@ -1,69 +1,223 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"os"
+	"strings"
 
 	c "github.com/ipfs/go-cid"
 
+	mb "github.com/multiformats/go-multibase"
 	mh "github.com/multiformats/go-multihash"
 )
 
+func usage() {
+	fmt.Fprintf(os.Stderr, "usage: %s [-b multibase-code] <fmt-str> <cid> ...\n\n", os.Args[0])
+	fmt.Fprintf(os.Stderr, "<fmt-str> is either 'prefix' or a printf style format string:\n%s", fmtRef)
+	os.Exit(1)
+}
+
+const fmtRef = `
+   %% literal %
+   %b multibase name 
+   %B multibase code
+   %v version string
+   %V version number
+   %c codec name
+   %C codec code
+   %h multihash name
+   %H multihash code
+   %L hash digest length
+   %m multihash encoded in base %b (with multibase prefix)
+   %M multihash encoded in base %b without multibase prefix
+   %d hash digest encoded in base %b (with multibase prefix)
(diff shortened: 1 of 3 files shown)


cid-fmt/main.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	for _, cidStr := range args[1:] {
+		base, cid, err := decode(cidStr)
+		if newBase != -1 {
+			base = newBase
+		}
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error: %s: %v\n", cidStr, err)
+			fmt.Fprintf(os.Stdout, "!INVALID_CID!\n")
+			// Don't abort on a bad cid
```
Stebalien: Hm. You sure about this? Shouldn't it at least return a non-zero exit status?

cid-fmt/main.go
```diff
@@ -0,0 +1,223 @@
...
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+			// An error here means a bad format string, no point in continuing
+			os.Exit(1)
+		}
+		fmt.Fprintf(os.Stdout, "%s\n", str)
+	}
+}
+
+func decode(v string) (mb.Encoding, *c.Cid, error) {
```
Stebalien: Can't we just use `cid.Decode`?

kevina: cid.Decode does to return the multibase used

cid-fmt/main.go
kevina: I can return non-zero exit status.  But if multiple CID are given it may be useful to try and convert the other arguments.

cid-fmt/main.go
Stebalien: Ah...

cid-fmt/main.go
Stebalien: If you want to try each one independently, you can call the command multiple times. As-is, you can dump the data to a file and not realize you have a problem till later.

Note: You don't have to exit immediately, my main concern is that the command should eventually exit with a non-zero status (the command was not a complete success).

kevina: @Stebalien I just pushed a commit to return non-zero status on any errors.

kevina:
@Stebalien note, that the test cases will need some fixing once https://github.com/multiformats/go-multibase/pull/19 is merged and the dependency updated.

Stebalien:
Understood (I merged this because I needed it and thought I had already merged it, it's *really* useful).

kevina:
(I am glad you found it useful.  :smile:  It will really help with the go-ipfs sharness tests.)

Issue #34, opened 2017-08-29
Provide Way of Representing CidV0 objects in an alternative base


kevina:
It may be useful to represent a CidV0 object in an alternative base.  For example if we need to include it in as the domain part of a URL.

I propose we provide allow a non non-standard encoding of:

`<multibase prefix><version><multihash>`

where version is always 0.

Related to https://github.com/ipfs/go-ipfs/issues/4143.

Note: I made this more complicated than it needs be, see this comment: https://github.com/ipfs/go-cid/issues/34#issuecomment-325868148

lgierth:
I've been thinking about this too, for the case where we want to represent a CIDv0 in base32 in a URL. The less obvious thing here is that we probably need to be able to recognize such CIDs and convert them back to CIDv0, so that DHT and Bitswap return the expected results.

Do you agree this is a concern? @whyrusleeping was skeptical when I first brought it up a while ago.

It could be done with a multicodec header that says "what's in here is CIDv0".

diasdavid:
> for the case where we want to represent a CIDv0 in base32 in a URL.

For those cases, we upgrade the CID to CIDv1 and base encoded it with base32.

kevina:
@lgierth 

> It could be done with a multicodec header that says "what's in here is CIDv0

Yes, and I think just setting the version number to "0" would work, hence my original proposal.

> For those cases, we upgrade the CID to CIDv1 and base encoded it with base32.

The ideas is we want to be able to retrieve a Cidv0 object using base32.  Upgrading the CID will not work in this case.

lgierth:
> The ideas is we want to be able to retrieve a Cidv0 object using base32. Upgrading the CID will not work in this case.

@diasdavid what Kevin and I are thinking about is the next step after receiving a gateway request with a CIDv0-in-CIDv1: content routing and bitswap, where we need the original CIDv0 because that's what provider records are published for, and what's used as a key in the blockstore of the provider nodes. It's kind of a matter of backward compatibility: there will always be nodes who continue to create CIDv0 stuff.

Stebalien:
@kevina I'd go a step further and start using the CIDv1 formatted CIDv0s by default. I believe that will actually be *less* work and, given that we're planning on switching to base32 by default anyways, shouldn't cause any additional breakage. Does that make sense or am I missing something?

kevina:
@Stebalien What exactly will be a CIDv1 formatted CIDv0?

[...]

PR #35, opened 2017-08-30
Fix StringOfBase to allow encoding CidV0 in alternative bases.

kevina:
See https://github.com/ipfs/go-cid/issues/34#issuecomment-325868148


kevina:
This is not going to happen (even though I still think it is a good idea) and is in some ways superseded by #60.  Closing.

PR #36, opened 2017-08-30
Add CID Prefix constructors.

Stebalien:
One can now generate a CID by calling:

```go
cid.NewPrefixV1(cid.DagCBOR, mh.SHA_256).Sum(data)
```

Lots of code was already doing this by manually constructing CID Prefixes but
that tends to be a bit verbose and error prone.


Stebalien:
cc @lmars

PR #40, opened 2018-02-22
Only allow crypto-safe hashes to be used in CIDs

whyrusleeping:
This is a WIP solution to https://github.com/ipfs/go-ipfs/issues/4371


validate.go
```diff
@@ -0,0 +1,63 @@
...
+	mh.SHAKE_256:    true,
+	mh.DBL_SHA2_256: true,
+	mh.KECCAK_224:   true,
+	mh.KECCAK_256:   true,
+	mh.KECCAK_384:   true,
+	mh.KECCAK_512:   true,
+	mh.ID:           true,
+
+	// insecure
+	mh.SHA1:    false,
```
Stebalien: This will, of course, break the git plugin.

validate.go
```diff
@@ -0,0 +1,63 @@
...
+	return false
+}
+
+func ValidateCid(c *Cid) error {
+	pref := c.Prefix()
+	if !IsGoodHash(pref.MhType) {
+		return ErrPossiblyInsecureHashFunction
+	}
+
+	if pref.MhType != mh.ID && pref.MhLength < minimumHashLength {
```
Stebalien: I'd rather check if the hash length >= the default. Alternatively, make `goodset` map hash function to minimum length (where we'd use -1 for ID).

validate.go
whyrusleeping: mmm, youre right.

whyrusleeping: As suggested by @magik6k we should probably set SHA1 as secure, but add the check described here: https://github.com/multiformats/go-multihash/issues/57 to detect insecure hashes.

validate.go
whyrusleeping: Hrm... I still want to allow some truncation. Ethereum truncates a lot of things down to 20 bytes (from 32). I can go with a mapping, but maybe default 20 and have a special cases for anything else?

validate.go
kevina: @whyrusleeping this would still break the git plugin, but only in the very rare case when the content so happens to hash to one of the insure hashes, or am I missing something?

validate.go
kevina: I think we should set this per hash as @Stebalien suggested to allow a little more flexibility.

Stebalien: @whyrusleeping WTF? That's 80 bits of security! No!

whyrusleeping: @Stebalien hold up, 8 * 20 is 160.

validate.go
whyrusleeping: yep

kevina: @whyrusleeping yep to what?  Am I missing something or will this still break the git plugin in the very rare case.

whyrusleeping: yes. This will still break the git plugin in that case.

validate.go
whyrusleeping: @Stebalien the whole point of multiformats in general is to allow users to use what fits their needs best. We should continue using 256bit hashes for our important things, but shouldnt begrudge users who want to use 160bit hashes. I'm primarily concerned about preventing pre-image attacks, general collision birthday problem attacks are still a concern, but less so for now.

validate.go
kevina: From IRC (slightly edits for clarity):

>**whyrusleeping**: juan wants this to make it into the next release
>*kevina*: whyrusleeping: I vote we allow "sha1" for now to avoid breaking the git plugin until we think of a better solution
>**whyrusleeping**: yeah, maybe.
>**whyrusleeping**: I think most of the concern we want to fix here is preimage attacks
>**whyrusleeping**: and theres no preimage attack for sha1
>**whyrusleeping**: just a collision
>**kevina** whyrusleeping, stebalien: so let's allow it with a note that it will eventually be marked insure once we implement something like /ipfs-insecure/
```

**Comment by Kubuxu** [comment]
@whyrusleeping I can work with this API if in future we want to swap this out for multihash-secruity-bits.

**Comment by kevina** [comment]
@Kubuxu can be save the commit "Enforce the security rules of crypto-safe hashes and min length" for a separate P.R.

**Comment by whyrusleeping** [comment]
@kevina no, its fine. I want this out of the way fast. Its blocking the next release which is severely overdue

**Comment by kevina** [comment]
@whyrusleeping I don't approve with the way this is being done then, as it is now it is impossible to create a CID with an insecure hash even when it might be useful.

**Comment by Kubuxu** [comment]
I would argue that as CID is Content Identifier, using unsafe hash is not longer identifier. @kevina could you share when it is useful?

**Comment by kevina** [comment]
@Kubuxu I can't think of anything right now, but it just feels like the wrong way to go about enforcing the rules for the "/ipfs" namespace, espacally if we want to eventually allow more liberal rules in a different namespace.  It has the same feel to me as the move to move to https every where, even when it is not really needed like over a local (and secure) network or the fact that you can't use ssh without encryption even though it could often be useful as a better version rsh over the local network.  If no one else agrees with me I will not object any further.

**Comment by kevina** [comment]
This will likely require a repo migration, to cleanly upgrade as I suggested in #4766.

**Comment by whyrusleeping** [comment]
we implemented a workaround in go-ipfs for now. This change could brick peoples repos if they already had bad hashes in them, so we removed the ability to create bad hashes, we're working on a migration to remove any existing bad hashes, and then once that is done, we can merge this.

**Comment by kevina** [comment]
@whyrusleeping @Stebalien is the only hold of this my repo migration?  Or is this blocked on anything else.  In particular how quickly can go-ipfs update a new `go-cid` version considering that it is used almost everywhere and will require a lot of package updates?

**Comment by Stebalien** [comment]
@kevina as far as I know, nothing is blocking a go-cid update. I can do it in a few hours with gx-workspace and a stress ball.

File: validate.go (hunk 2 above)
Comment (Stebalien) [reply validate.go]: So, I've come to agree on this. It still annoys me, but we don't have much of a choice.

---

If we ever need to set different minimum lengths for different hashes, we can change that later.

File: validate.go [hunk 3]
```diff
@@ -0,0 +1,61 @@
...
+	mh.SHA3_512:     true,
+	mh.SHAKE_256:    true,
+	mh.DBL_SHA2_256: true,
+	mh.KECCAK_224:   true,
+	mh.KECCAK_256:   true,
+	mh.KECCAK_384:   true,
+	mh.KECCAK_512:   true,
+	mh.ID:           true,
+
+	mh.SHA1: true, // not really secure but still useful
```
Stebalien: Could we restrict this to git only somehow?

validate.go
Stebalien: Also, users can implement custom DAGServices for validating additional properties (we may want won for unixfs that has stricter requirements).

validate.go
kevina: I don't necessary agree with this unless we are 100% sure SHA-1 are not used anywhere outside of git.  Otherwise we could break things, badly.

kevina:
@whyrusleeping @Kubuxu is it okay if I take over this p.r.?  For now I pushed my work on #50.

validate.go
Stebalien: The only reason we want to call SHA1 a "secure" hash because we want git compatibility. SHA1 is not a secure hash function. I *know* there are no other existing IPLD formats using SHA1 by default. If users have *chosen* to use SHA1 with IPLD, well, I'm not exactly sympathetic.

If we find some existing merkledag format that uses SHA1 and want to add compatibility, we can add that to the list later.

---

Note: I'm fine pushing this patch through as-is and fixing this later as this patch *is* a strict improvement.

lidel:
This PR never landed and got out of date,
I am closing it as it is most likely easier to create new one if we ever want to move beyond current solution (blocking at bockstore instead in go-cid – continued in https://github.com/ipfs/go-ipfs/issues/4371#issuecomment-400031808 ++

PR #42, opened 2018-03-13
Add Keccak and SHA hashes

gtfierro:
I'm working on an IPFS-based project that requires use of some of these alternative hash functions. I've moved a few over from the [multihash table](https://github.com/multiformats/multihash/blob/master/hashtable.csv).

Specifically, I've added SHA1, SHA2-256, SHA2-512, Keccak-224, Keccak-256,
Keccak-384 Keccak-512. Have not added tests at this point but I'm not sure what the most effective tests would be. Let me know what I can do to help!

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -77,6 +77,15 @@ const (
 	BitcoinTx          = 0xb1
 	ZcashBlock         = 0xc0
 	ZcashTx            = 0xc1
+
+	SHA1     = 0x11
+	SHA2_256 = 0x12
+	SHA2_512 = 0x13
+
+	Keccak224 = 0x1a
+	Keccak256 = 0x1b
+	Keccak384 = 0x1c
+	Keccak512 = 0x1d
 )
 
 // Codecs maps the name of a codec to its type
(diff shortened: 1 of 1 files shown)


Stebalien:
Those tables are actually for IPLD object formats. CIDs are really just multihashes wrapped in some extra metadata so all of those hash formats are already supported. You can specify the multihash codec type when creating the CID.

gtfierro:
The specific issue I'm trying to fix is [go-ipfs-api](https://github.com/ipfs/go-ipfs-api) uses go-cid but only allows the specification of formats that are listed in the `Codecs` map. I could change those libraries so that they accept the integer identifiers as well as the strings, but that was more of a breaking change than just adding those values here. I'll just move ahead with a fork of go-ipfs-api for now. thanks!

Stebalien:
I see. That's because kaccak-256 is not a block format, it's a hash algorithm. Block (IPLD) formats include things like DagCBOR, DatPB, DagRaw, DagEth, DagGitRaw, etc...

You're actually looking for the `--mhtype` option (multihash type).

gtfierro:
For my particular application I need the block bytes to be the keccak-256 hash of the content so that I can do content-addressing across multiple storage providers (not just IPFS). I'll move ahead with a fork for now.

Thanks!

Stebalien:
So you have a hash of a hash? CID -> keccac-256-hash -> data? Can you not just convert the keccac-256 hash to a CID. That's how we handle Git and Eth objects. IPLD was specifically *designed* to handle arbitrary content addressing schemes.

Trust me, you're going down the wrong path here.

gtfierro:
You were totally right. I dug into it a bit more and found that the Raw object format gives me the intended result, without me needing to change any of the internals of these packages. Thanks for your help!

Stebalien:
Np. Didn't want you going down that rabbit hole :smile:.

However, if you're pointing to Keccac hashes that, in turn, point to something else, you may be able to avoid the indirection by turning the Keccac hashes into CIDs. That is:

```go
import (
	mh "github.com/multiformats/go-multihash"
	cid "github.com/ipfs/go-cid"
)

// ...

hash, err := mh.Encode(bareKeccacHash, mh.KECCAK_256)
yourCid := cid.NewCidV1(cid.Raw, hash)
```

PR #43, opened 2018-04-20
support nil CIDs when checking equality

Stebalien:
Treat nil CIDs like NaN. They never equal anything.

(also, short-circuit equality check for self == self)

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -371,6 +371,12 @@ func (c *Cid) bytesV1() []byte {
 // In order for two Cids to be considered equal, the
 // Version, the Codec and the Multihash must match.
 func (c *Cid) Equals(o *Cid) bool {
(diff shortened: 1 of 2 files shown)


Kubuxu:
@Stebalien what is the reason behind this?

Stebalien:
@Kubuxu 

1. The identity short-circuit may help a bit with performance.
2. I'd like to avoid panicing when handling nil CIDs (the drawback is that this could hide bugs).

I remember wanting this somewhere but I can't remember where exactly.

kevina:
Closing.  The motivation is unclear and in any case it will become irrelevant once we switch to a string backing.

Stebalien:
Yeah, we actually get this support out of the box...

Issue #38, opened 2017-10-20
Suggestion: Change CID type so it can be passed around directly and not by pointer


kevina:
Right now the Cid type is:
```go
type Cid struct {
    version uint64
    codec   uint64
    hash    mh.Multihash
}
```
The internals are not exposed but it is expected to be passed around by pointer which in a way exposes some of the implementation details.  For example in #3 I suggested we use a more compact structural representation and I think @Stebalien is suggesting to a serialized string instead.  The best representation may be debatable, but either change will require that the Cid no longer be passed around by a pointer.  We may even decide that a Cid is better represented by an interface and some point.

So for now may I suggestion that we start passing around the Cid type directly and change the representation to.

```go
type Cid struct {
    *cid
} 
type cid struct {
    version uint64
    codec   uint64
    hash    mh.Multihash
}
```
This will then give us the freedom to try different internal representations without making requiring any API changes. 

@whyrusleeping @Stebalien thoughts?

Stebalien:
From my experience today trying to replace `Multiaddr` with a bare type, this will be *fun* as we probably rely on the fact that Cids are nullable all over the place (although I agree we should probably do this anyways).

We could also just go straight to an interface.

Note: For my usecase, it's actually fine to continue passing it around as a pointer as long as I can cheaply extract some key to use in a map (either by dereferencing it or pulling out some field).

kevina:
> this will be _fun_

If we decide to go though with this I can likely do most of the work and probably in a day or two.  I will need someone to help me with using gx-workspace, though, for handling the dependency updates.

> probably rely on the fact that Cids are nullable all over the place

Maybe, in the cases where this is true we can use a pointer or we can have a special "empty" CID value.  We won't know how bad this is until we try.

> We could also just go straight to an interface.

I would be against this, it adds an extra level of indirection that I was trying to get ride of.  I also think it will be overkill in this case.

PR #47, opened 2018-05-17
use string instead of struct backing

dignifiedquire:
cc @Stebalien 

ref #38

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -423,7 +423,7 @@ func (c Cid) MarshalJSON() ([]byte, error) {
 
 // KeyString casts the result of cid.Bytes() as a string, and returns it.
 func (c Cid) KeyString() string {
-	return string(c.Bytes())
+	return string(c)
 }
 
 // Loggable returns a Loggable (as defined by
--- a/cid.go
+++ b/cid.go
@@ -126,22 +126,28 @@ var CodecToStr = map[uint64]string{
 // They exist to allow IPFS to work with Cids while keeping
 // compatibility with the plain-multihash format used used in IPFS.
 // NewCidV1 should be used preferentially.
-func NewCidV0(mhash mh.Multihash) *Cid {
-	return &Cid{
-		version: 0,
-		codec:   DagProtobuf,
-		hash:    mhash,
-	}
+func NewCidV0(mhash mh.Multihash) Cid {
+	return newCid(0, DagProtobuf, mhash)
 }
 
 // NewCidV1 returns a new Cid using the given multicodec-packed
 // content type.
-func NewCidV1(codecType uint64, mhash mh.Multihash) *Cid {
-	return &Cid{
-		version: 1,
-		codec:   codecType,
-		hash:    mhash,
+func NewCidV1(codecType uint64, mhash mh.Multihash) Cid {
+	return newCid(1, codecType, mhash)
+}
+
+func newCid(version, codecType uint64, mhash mh.Multihash) Cid {
+	hashlen := len(mhash)
+	// two 8 bytes (max) numbers plus hash
+	buf := make([]byte, 2*binary.MaxVarintLen64+hashlen)
+	n := binary.PutUvarint(buf, version)
+	n += binary.PutUvarint(buf[n:], codecType)
+	cn := copy(buf[n:], mhash)
+	if cn != hashlen {
+		panic("copy hash length is inconsistent")
 	}
+
+	return Cid(buf[:n+hashlen])
 }
 
 // NewPrefixV0 returns a CIDv0 prefix with the specified multihash type.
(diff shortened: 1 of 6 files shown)


Stebalien:
I like this. However, it's a massively breaking change so we'll have to be careful (I've received a lot of push-back from @whyrusleeping when trying to do this kind of thing in the past). Also, this makes extracting a multihash from a CID allocate (could be fixed by using a string internally) and means that constructing a CID involves copying bytes. We can fix the former by storing multihashes as strings (https://github.com/multiformats/go-multihash/issues/29) and the latter by adding a function to multihash that writes to a writer (e.g., a string builder) which we'll probably want anyways.

A less breaking-change version would be to use a pointer to a string (or a pointer to, e.g., `struct Cid{ string }`). However, that will require an additional allocation (although the `KeyString` method becomes trivial which is my personal goal).

---

Performance wise, implementing the varint skipping code manually may be faster. That is, to skip two varints, count two bytes with the high-order bit unset.

---

Regardless, I think we should wait for the libp2p refactor + dht fixes + log fixes + everything else that is currently blocked.

Personally, I'd *like* to just break things and fix this all at once but this really is a massively breaking change. All `cid == nil` checks now no longer work.

whyrusleeping:
@dignifiedquire please familiarize yourself with the other attempts to do similar things, primarily: https://github.com/multiformats/go-multiaddr/pull/62

dignifiedquire:
@whyrusleeping I did read through that discussion, but didn't see that there was a final decision made on the direction to go, so this is based on a conversation I had with @Stebalien afterwards. 

I am happy to adjust if you and @Stebalien get to an agreement on which version exactly you want to go for in these, just let me know.

cid.go
```diff
@@ -168,15 +174,29 @@ func NewPrefixV1(codecType uint64, mhType uint64) Prefix {
...
-	version uint64
-	codec   uint64
-	hash    mh.Multihash
+// Byte layout: [version, codec, multihash]
+// - version uvarint
+// - codec uvarint
+// - hash mh.Multihash
+type Cid string
+
+var EmptyCid = Cid(string([]byte{}))
```
kevina: Why `string([]byte{})` and not just `""`.

kevina: I would also make this a constant so we have:
```
const EmptyCid = Cid("")
```

kevina: Also. I find `EmptyCid` a bit tedious of a name.  How out just `Ni`"?  When used outside this package it would be `cid.Nil`.

cid.go
```diff
@@ -350,30 +368,20 @@ func (c *Cid) Bytes() []byte {
...
 
 // Equals checks that two Cids are the same.
 // In order for two Cids to be considered equal, the
 // Version, the Codec and the Multihash must match.
-func (c *Cid) Equals(o *Cid) bool {
-	return c.codec == o.codec &&
-		c.version == o.version &&
-		bytes.Equal(c.hash, o.hash)
+func (c Cid) Equals(o Cid) bool {
+	// TODO: can we use regular string equality?
```
kevina: > can we use regular string equality?

Yes.

cid.go
```diff
@@ -168,15 +174,29 @@ func NewPrefixV1(codecType uint64, mhType uint64) Prefix {
...
 // a multicodec-packed content type) and a Multihash.
-type Cid struct {
-	version uint64
-	codec   uint64
-	hash    mh.Multihash
+// Byte layout: [version, codec, multihash]
+// - version uvarint
+// - codec uvarint
+// - hash mh.Multihash
+type Cid string
```
kevina: Maybe we should make this `type Cid struct{ string }` to ensure the Cid type is always valid.

cid_test.go
```diff
@@ -383,3 +373,39 @@ func TestFromJson(t *testing.T) {
...
+	count := 0
+	for i := 0; i < b.N; i++ {
+		count += len(cid.String())
+	}
+	if count != 49*b.N {
+		b.FailNow()
+	}
+}
+
+// making sure we don't allocate when returning bytes
```
kevina: But we do now and can't avoid it.  I take it `BenchmarkBytesV1` is leftover from the first attempt with used []byte as the representation.

Stebalien:
@dignifiedquire for context, this is moving forward do to:

#70, the fact that by-value CIDs will work better with refmt, and the fact that we're reworking a ton of CID stuff *anyways* for the base32 endeavor.

dignifiedquire:
@stebalien thanks for the heads up, I was following the other thread. Let me know if you need anything from me, otherwise feel free to take the code and use what you can, and throw away what you don’t like :)

kevina:
Closing in favor of #71

PR #48, opened 2018-06-04
Factor out Prefix helper functions from go-ipfs/merkledag

kevina:


prefix.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// V0CidPrefix returns a prefix for CIDv0
+func V0CidPrefix() Prefix { return v0CidPrefix }
+
+// V1CidPrefix returns a prefix for CIDv1 with the default settings
+func V1CidPrefix() Prefix { return v1CidPrefix }
+
+// PrefixForCidVersion returns the Protobuf prefix for a given CID version
+func PrefixForCidVersion(version int) (Prefix, error) {
```
Stebalien: IMO, this is *way* too protobuf specific to go in go-cid.

kevina:
> Do the NewPrefixV0 and NewPrefixV1 functions not work for you?

They would if they existed, but I only fond NewCidV0 and NewCidV1.

prefix.go
kevina: Fair enough.  Will change to make them more general.

Stebalien:
> They would if they existed, but I only fond NewCidV0 and NewCidV1.

They're in cid.go.

kevina:
> They're in cid.go.

Oops.  Forgot to update my local git copy.  Sorry about that.

Closing.

Stebalien:
Np. I've done that countless times.

PR #51, opened 2018-06-25
Tweak parameters of NewPrefix functions.

### LINKED ISSUE #49: NewPrefixV0 should not take a multihash
They is only one legal multihash in CidV0 so requiring one is redundant.

kevina:
I don't think these two functions are really used yet, so this shouldn't break anything.

`NewPrefixV0` doesn't take any parameters, since there is only one possible legal combination.  `NewCidV1` is expanded to also take the multihash length.  Even though the length 99% of the time will be `-1` it still makes sense to require it.  This gives the function the opportunity to convert the length value of `-1` to the default length.  The callee of the function is already likely to have the length value and needs to do something with it.  With out being able to pass it in the most logical think to do would be to set it afterwards, for example:
```
  func foo(..., mh uint64, mhlen int) { 
    prefix := cid.NewPrefixV1(..., mh)
    prefix.MhLength = mhlen
  }
```
which doesn't allow NewPrefix to set the value to the default length.

Closes #49.

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -145,10 +145,10 @@ func NewCidV1(codecType uint64, mhash mh.Multihash) *Cid {
 }
 
 // NewPrefixV0 returns a CIDv0 prefix with the specified multihash type.
-func NewPrefixV0(mhType uint64) Prefix {
+func NewPrefixV0() Prefix {
 	return Prefix{
-		MhType:   mhType,
-		MhLength: mh.DefaultLengths[mhType],
+		MhType:   mh.SHA2_256,
+		MhLength: mh.DefaultLengths[mh.SHA2_256],
 		Version:  0,
 		Codec:    DagProtobuf,
 	}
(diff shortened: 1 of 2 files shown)


kevina:
@Stebalien do you think we could get this change in ASAP?  Right now it doesn't look like these functions are used.  Once they become used this change will become more difficult.

Stebalien:
So, aren't we planning on replacing these functions with a `Format` struct and a `NewFormat` function?

Stebalien:
That is, I'd rather deprecate and remove these functions and move to that than to break things.

Stebalien:
Possible way forward:

1. Merge #52 first.
2. Make everything that uses these functions use those.
3. Deprecate these functions.
4. Remove at our leisure.

Basically, changing the signature will be the same amount of work as deprecating and removing.

kevina:
@Stebalien 

Not merging this P.R. is partly blocking #50.

A grep of the entire source tree found that these functions are used in only one place outside of go-cid in `go-libp2p-routing-helpers` in `parallel_test.go`.

#52 will take awhile to review and get in, this is a very simple change.

My plan was to get #40/#50 in before #52.

As far work, all that needs to be done is to just merge this P.R.  We don't need to GX publish just yet.  If this gets GX published before #52 gets in fixing the signature in the one place is not a lot of work compared to the updating all the places that use `go-cid`.

Stebalien:
> As far work, all that needs to be done is to just merge this P.R. We don't need to GX publish just yet. If this gets GX published before #52 gets in fixing the signature in the one place is not a lot of work compared to the updating all the places that use go-cid.

That'll still break anyone calling `go test ./...`. While we *use* gx, we try not to rely on it.

Also, there are other users:

* go-ipfs-blockstore
* Bit-Nation/panthalassa
* mgoelzer/libp2p-chat-app
* mikiquantum/rock-paper-scisors-demo
* ...

If an interface is broken and we'll want to break it eventually, I'm usually all for breaking it sooner rather than later. However, given that we want to simply *replace* these functions, I'd rather leave them as-is (they still work, even if they allow invalid states) and then replace and deprecate them.

kevina:
Closing in favor of #53.

PR #52, opened 2018-06-28
add some helpers for creating new CIDs

Stebalien:
Unlike Prefix, we can extend continue to extend this helper as the options grow.

Issue #23, opened 2017-06-20
Decide how to handle -1 in Prefix


Stebalien:
Currently, `Prefix.MhLength` is an `int` and `-1` can be (and is) used to mean "default length". Unfortunately, this means:

1. `cid1.Bytes() == cid2.Bytes()` does not imply `cid1.Prefix() == cid2.Prefix()`.
2. `Prefix.Bytes()` is broken.

Solutions:

1. Make it a uint64, provide some convenience constructors constructors (e.g. `func V1Prefix(codec uint64, mhType uint64) Prefix`). This will break things.
2. Fix `Prefix.Bytes()` and provide an `Equals` method (less convenient in the long run).

Thoughts?

kevina:
@Stebalien 

I think the real problem is we are using Prefix for two different things which I think should be separated.

(1) To get information about an existing CD
(2) To create a new CID

I think we should separate these two concerns.  This is espacally relevant now as I am trying to think how to add support for automatically creating Identity hashes.  I am tempted to hack on support with the existing Prefix structure but will only make things worse.

Instead I propose we use two different structures one to get information about a Cid and another to be used to actually generate a Cid, the first will have a Bytes() method but no Sum() method, and a unsigned MhLength; the latter will not have a Bytes() method, but will have a Sum() method, and a signed MhLength.  The latter will also be extended to support creating identity hashes.

The only creation is what to call to two structures, should we use `CidInfo` and `Prefix` (where the first is for getting info on a Cid and the later for creating them); or maybe `Prefix` an `CidOpts`; or maybe rename both to `CidInfo` and `CidOpts`; or maybe something else as I am horrible and thinking of good names for things.

Thoughts?

kevina:
Unless I am missing something obvious (always possible) getting the prefix of a Cid() is the only way to get the Version of the cid, although we could add a Version() method fairly easily.

Other than that I am not sure and would need to check everywhere Prefix is used.

schomatis:
For me, the most confusing part is that I don't find the `Prefix` in the [CID](https://github.com/ipld/cid/search?q=prefix) specification so I don't know what to expect of this structure. From the code it seems that `Prefix` has pretty much the same information as `Cid` minus the digest. It seems to be used for this double function as @kevina mentioned,

> (1) To get information about an existing CD
> (2) To create a new CID

[...]

PR #53, opened 2018-06-29
Create new Builder interface for creating CIDs.

kevina:
This is my proposed alternative to #52 

My making `Format` an interface we can easily extend it to provide additional functionality outside of the `go-cid` package.  In addition the current `Prefix` implemented the `Format` interface so to take advantage of the added functional all we need to is is change the signature of functions that take a `cid.Prefix` to `cid.Format`.

I specifically did not add support for automatically creating identity hashes as that can be added later, and possible outside of `go-cid`.  The code could look something like:
```go
  struct FormatWithInlining {Base Format, Limit int}

  func WithInlinling(base Format) {
    return FormatWithInlining{Base: base, Limit: DefaultInlineLimit}
  }

  func (f FormatWithInlining) Sum(data []byte) (*Cid, error) {
    if len(data) <= f.Limit {/* use id hash*/}
    else {return f.Base.Sum(data)}
  }
```

At some point in the future we can remove NewPrefixV0, NewPrefixV1 and Prefix.Sum.  We should also consider making the MhLength field an unsigned integer to disallow -1 as a value.

Some additional nodes:  I find the field names `MhType` and `MhLength` crypt and think `HashFun` and `HashLen` are better and easier to understand names so I used those in this structure.  If we are dead set on `MhType` and `MhLength` I can change it back.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -4,23 +4,23 @@ import (
 	mh "github.com/multiformats/go-multihash"
 )
 
-type Format interface {
+type Builder interface {
 	Sum(data []byte) (*Cid, error)
 	GetCodec() uint64
-	WithCodec(uint64) Format
+	WithCodec(uint64) Builder
 }
 
-type FormatV0 struct{}
+type V0Builder struct{}
 
-type FormatV1 struct {
+type V1Builder struct {
 	Codec   uint64
 	HashFun uint64
 	HashLen int // HashLen <= 0 means the default length
 }
 
-func PrefixToFormat(p Prefix) Format {
+func PrefixToBuilder(p Prefix) Builder {
 	if p.Version == 0 {
-		return FormatV0{}
+		return V0Builder{}
 	}
 	mhLen := p.MhLength
 	if p.MhType == mh.ID {
@@ -29,7 +29,7 @@ func PrefixToFormat(p Prefix) Format {
 	if mhLen < 0 {
 		mhLen = 0
 	}
-	return FormatV1{
+	return V1Builder{
 		Codec:   p.Codec,
 		HashFun: p.MhType,
 		HashLen: mhLen,
@@ -40,7 +40,7 @@ func (p Prefix) GetCodec() uint64 {
 	return p.Codec
 }
 
-func (p Prefix) WithCodec(c uint64) Format {
+func (p Prefix) WithCodec(c uint64) Builder {
 	if c == p.Codec {
 		return p
 	}
@@ -51,7 +51,7 @@ func (p Prefix) WithCodec(c uint64) Format {
 	return p
 }
 
-func (p FormatV0) Sum(data []byte) (*Cid, error) {
+func (p V0Builder) Sum(data []byte) (*Cid, error) {
 	hash, err := mh.Sum(data, mh.SHA2_256, -1)
 	if err != nil {
 		return nil, err
@@ -59,18 +59,18 @@ func (p FormatV0) Sum(data []byte) (*Cid, error) {
 	return NewCidV0(hash), nil
 }
 
-func (p FormatV0) GetCodec() uint64 {
+func (p V0Builder) GetCodec() uint64 {
 	return DagProtobuf
 }
 
-func (p FormatV0) WithCodec(c uint64) Format {
+func (p V0Builder) WithCodec(c uint64) Builder {
 	if c == DagProtobuf {
 		return p
 	}
-	return FormatV1{Codec: c, HashFun: mh.SHA2_256}
+	return V1Builder{Codec: c, HashFun: mh.SHA2_256}
 }
 
-func (p FormatV1) Sum(data []byte) (*Cid, error) {
+func (p V1Builder) Sum(data []byte) (*Cid, error) {
 	mhLen := p.HashLen
 	if mhLen <= 0 {
 		mhLen = -1
@@ -82,11 +82,11 @@ func (p FormatV1) Sum(data []byte) (*Cid, error) {
 	return NewCidV1(p.Codec, hash), nil
 }
 
-func (p FormatV1) GetCodec() uint64 {
+func (p V1Builder) GetCodec() uint64 {
 	return p.Codec
 }
 
-func (p FormatV1) WithCodec(c uint64) Format {
+func (p V1Builder) WithCodec(c uint64) Builder {
 	p.Codec = c
 	return p
 }
--- a/builder.go
+++ b/builder.go
@@ -13,9 +13,9 @@ type Builder interface {
 type V0Builder struct{}
 
 type V1Builder struct {
-	Codec   uint64
-	HashFun uint64
-	HashLen int // HashLen <= 0 means the default length
+	Codec    uint64
+	MhType   uint64
+	MhLength int // MhLength <= 0 means the default length
 }
 
 func (p Prefix) GetCodec() uint64 {
@@ -49,15 +49,15 @@ func (p V0Builder) WithCodec(c uint64) Builder {
 	if c == DagProtobuf {
 		return p
 	}
-	return V1Builder{Codec: c, HashFun: mh.SHA2_256}
+	return V1Builder{Codec: c, MhType: mh.SHA2_256}
 }
 
 func (p V1Builder) Sum(data []byte) (*Cid, error) {
-	mhLen := p.HashLen
+	mhLen := p.MhLength
 	if mhLen <= 0 {
 		mhLen = -1
 	}
-	hash, err := mh.Sum(data, p.HashFun, mhLen)
+	hash, err := mh.Sum(data, p.MhType, mhLen)
 	if err != nil {
 		return nil, err
 	}
--- a/builder.go
+++ b/builder.go
@@ -18,24 +18,6 @@ type V1Builder struct {
 	HashLen int // HashLen <= 0 means the default length
 }
 
-func PrefixToBuilder(p Prefix) Builder {
-	if p.Version == 0 {
-		return V0Builder{}
-	}
-	mhLen := p.MhLength
-	if p.MhType == mh.ID {
-		mhLen = 0
-	}
-	if mhLen < 0 {
-		mhLen = 0
-	}
-	return V1Builder{
-		Codec:   p.Codec,
-		HashFun: p.MhType,
-		HashLen: mhLen,
-	}
-}
-
 func (p Prefix) GetCodec() uint64 {
 	return p.Codec
 }
--- a/format.go
+++ b/format.go
@@ -1,0 +1,55 @@
+package cid
+
+import (
+	mh "github.com/multiformats/go-multihash"
+)
+
+type Format interface {
+	Sum(data []byte) (*Cid, error)
+}
+
+type FormatV0 struct{}
+
+type FormatV1 struct {
+	Codec   uint64
+	HashFun uint64
+	HashLen int // HashLen <= 0 means the default length
+}
+
+func PrefixToFormat(p Prefix) Format {
+	if p.Version == 0 {
+		return FormatV0{}
+	}
+	mhLen := p.MhLength
+	if p.MhType == mh.ID {
+		mhLen = 0
+	}
+	if mhLen < 0 {
+		mhLen = 0
+	}
+	return FormatV1{
+		Codec:   p.Codec,
+		HashFun: p.MhType,
+		HashLen: mhLen,
+	}
+}
+
+func (p FormatV0) Sum(data []byte) (*Cid, error) {
+	hash, err := mh.Sum(data, mh.SHA2_256, -1)
+	if err != nil {
+		return nil, err
+	}
+	return NewCidV0(hash), nil
+}
+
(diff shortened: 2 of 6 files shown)


Stebalien:
> At some point in the future we can remove NewPrefixV0, NewPrefixV1 and Prefix.Sum

I'd leave `Prefix.Sum`. Being able to treat a prefix as a format is useful. However, we'll need to be careful to ignore the hash length for ID hashes (may need to fix that?).

> Some additional nodes: I find the field names MhType and MhLength crypt and think HashFun and HashLen are better and easier to understand names so I used those in this structure. If we are dead set on MhType and MhLength I can change it back.

I don't care much either way. However, I'd slightly prefer remaining consistent with `Prefix`.

format.go
```diff
@@ -0,0 +1,55 @@
...
+
+type FormatV0 struct{}
+
+type FormatV1 struct {
+	Codec   uint64
+	HashFun uint64
+	HashLen int // 0 (or -1 for compatibility) means the default length
+}
+
+func PrefixToFormat(p Prefix) Format {
```
Stebalien: Not sure if we really need this given that all `Prefix`s are `Format`s.

format.go
```diff
@@ -0,0 +1,55 @@
...
+type Format interface {
+	Sum(data []byte) (*Cid, error)
+}
+
+type FormatV0 struct{}
+
+type FormatV1 struct {
+	Codec   uint64
+	HashFun uint64
+	HashLen int // 0 (or -1 for compatibility) means the default length
```
Stebalien: I'd just say `HashLen` <= 0 means "default".

format.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	return NewCidV0(hash), nil
+}
+
+func (p FormatV1) Sum(data []byte) (*Cid, error) {
+	mhLen := p.HashLen
+	if mhLen == 0 {
+		mhLen = -1
+	}
+	hash, err := mh.Sum(data, p.HashFun, mhLen)
```
Stebalien: Should check if we're using the ID hash and verify that the hash length is <= 0

format.go
kevina: The idea was that we might want to deprecate Prefix.Sum.  This way we we can format MhLength from being -1, which is a source of confusion.

Stebalien: We can just check that on Sum (after a deprecation period). That is, I'd like to move away from having `MhLength == -1` (except for ID hashes, maybe). However, this is *only* an issue right now because the best way to create a CID is to construct a Prefix and then call Sum (and lazy programmers just set `MhLength = -1`).

Basically, `Prefix.Sum` isn't the problem (and is quite useful). It's `Prefix.MhLength == -1`.

kevina: I would go one step further and not add an exception for ID hashes.  That is a round trip from Prefix.Sum(data) and back to Prefix should create identical results.  If you want to create an ID hash use the new FormatV1 struct.

format.go
Stebalien: Fair enough.

kevina:
>> Some additional nodes: I find the field names MhType and MhLength crypt and think HashFun and HashLen are better and easier to understand names so I used those in this structure. If we are dead set on MhType and MhLength I can change it back.

> I don't care much either way. However, I'd slightly prefer remaining consistent with Prefix.

I find `FormatV1{Codec: ..., HashFun: ...}` clearer then `FormatV1{Codec: ..., MhType: ...}`, espacally to someone who doesn't know what `Mh` is an abbreviation for.

@Mr0grog any opinions on this, you where pretty vocal abort short names in https://github.com/ipfs/go-ipfs/issues/5056

kevina:
@Stebalien I updated the depreciated notes so that only setting `MhLength` to -1 in `Prefix` is deprecated and not the Sum method.  I also think we should consider changing `MhLength`  to an unsigned integer at somepoint to prevent any confusion.

`PrefixToFormat` might still be useful in the case of identify hashes because I believe fairly strongly that the round trip from `Prefix.Sum(data)` and back to `Prefix` should always create the same prefix.  This limits the usefulness of Prefix for identity hashes.  However, if you think the presence of that function is confusing I can remove it.

Mr0grog:
> I find `FormatV1{Codec: ..., HashFun: ...}` clearer then `FormatV1{Codec: ..., MhType: ...}`… @Mr0grog any opinions on this, you where pretty vocal abort short names

Ha! I’m not very familiar with this package at all, so take my commentary with a grain of salt. Knowing nothing else, I’d agree: `HashFun` and `HashLen` are clearer to me than `MhType` and `MhLength` (though I’m sure it will surprise nobody that I’d prefer `HashFunction`/`HashLength` or `MultihashType`/`MultihashLength` 😜).

BUT, I am sympathetic to @Stebalien’s point on consistency with `Prefix`. I’m just not entirely sure on the intended context or purpose of `Format` vs. `Prefix` (they both contain all the same info, just differently arranged and with different levels of explict-ness), so I’m not sure how to weigh that concern.

Either way, I think it would be helpful to clarify via comments that the field is intended to be [a hash function from the `multihash` package](https://github.com/multiformats/go-multihash/blob/a813e7a829f45ce9c38cb2e7326d1bbfb55ae4a8/multihash.go#L38-L64). Since that package doesn’t export a special type we can mark things with and we just use a `uint64` everywhere instead, it’s up to the comments/docs or the argument name to explain it (`MhType` does this if only someone knows what the `Mh` means).

(Sorry for the slow reply; just got back from a few days away.)

Stebalien:
> BUT, I am sympathetic to @Stebalien’s point on consistency with Prefix. I’m just not entirely sure on the intended context or purpose of Format vs. Prefix (they both contain all the same info, just differently arranged and with different levels of explict-ness), so I’m not sure how to weigh that concern.

Prefix includes information on existing CIDs. Format is for creating new CIDs. We currently use Prefix for both but that leads to some odd cases (e.g, the user wants to specify "just use the default hash length" but a valid prefix technically needs to *explicitly* specify the hash length).

Mr0grog:
>  We currently use Prefix for both but that leads to some odd cases

Ah, ok. Is this mainly so code that deals with it never has to look up the right hash length the way `Format.Sum()` does? I guess I was looking at the code like:

```go
metadata := cid.NewPrefixV1(cid.DagCBOR, mh.SHA2_256)
// or if Prefix-using functions could handle default length lookups when 0, or there was an accessor for hash length:
// metadata := cid.Prefix{Version: 1, Codec: cid.DagCBOR, MhType: mh.SHA2_256}
id := metadata.Sum(bytesOfContent)

// vs.

metadata := cid.FormatV1{Codec: cid.DagCBOR, HashFun: mh.SHA2_256}
id := metadta.Sum(bytesOfContent)
```

…which didn’t feel like it was getting us much,  but I also don’t know what "additional functionality" we are heading towards. If there’s a lot on the table, I can see how all the empty/optional fields a struct affords that Go’s lack of optional arguments doesn’t would be useful.

Since these two structs are so similar, this definitely feels like it needs some documentation pointing the way (both text in the package overview saying you should use the `Format` struct to create a CID for some bytes vs. `NewCidVx()` if you already have a hash and some docs in `format.go`), and maybe an example in the test file. I don’t think I’ll be alone in not totally understanding this.

**Anyway, on the original point:** all that does make me feel like having the names match is pretty valuable in this case (unless we might be comfortable changing `Prefix` to use different field names). `MhType` and `HashFun` would be representing the exact same thing and for pretty much the same purpose in both places. That’s a good reason to keep the names consistent, even if I’m not a big fan of the names themselves :)

Stebalien:
> …which didn’t feel like it was getting us much, but I also don’t know what "additional functionality" we are heading towards.

So, there were two concrete issues:

* https://github.com/ipfs/go-cid/issues/23
* Additional options like when to inline the data into the CID (i.e., when to *not* use the specified hash function and to instead use the identity hash function).

Basically, `Prefix` is a valid `Format` that can produce CIDs that all have the same `Prefix`. `Format` allows us to define a single CID formatter that can spit out CIDs with different formats.

---

Actually, @kevina, that reminds me of something. This version hardcodes the codec and doesn't provide a way to change it. That could be a problem for raw nodes etc. I'm thinking we may want something that takes the Codec in the `Sum` function. That is, `Sum(codec int, data []byte) (*cid.Cid, error)`. The format would then dictate all the "choices" (CIDv0 versus CIDv1, inline data, hash function, etc.) and could even choose these based on the codec. Thoughts? Have you taken a stab at using this interface with your inline CID patches?

Mr0grog:
> #23

Ah, thanks for this; it explains a good bit of background.

> `Format` allows us to define a single CID formatter that can spit out CIDs with different formats.

Oh! I don’t think I would have expected that from the name.

Stebalien:
> Oh! I don’t think I would have expected that from the name.

Any suggestions? We've been through quite a few names but haven't found one we're completely satisfied with.

Mr0grog:
Hmm… `Formatter`, `CidOptions`, `FormatOptions`, `Builder`, `CidBuilder` (it strikes me that the point and purpose of this is not too far from the builder pattern)? Best I have at the moment.

kevina:
> Actually, @kevina, that reminds me of something. This version hardcodes the codec and doesn't provide a way to change it.

Your right.  I added methods to do so and tested this now with `go-ipfs`.

kevina:
I think `Format` is not such a nice name for this?  @Stebalien is there a reason for this name.  I think maybe `Opts` will be better (`CidOpts` would be redundant).

For one thing I think renaming `Prefix` to `Format` in field names within `go-ipfs` would be confusing, we could use `CidFormat` but I think naming this to `Opts` and the field `CidOpts` would be better.  What do you think?  Note, I am somewhat against the longer `Options` name as I find `Opts` a fairly easy to recognized abbreviation.

kevina:
@Stebalien @Mr0grog after thhinking about it I think `Builder` will be a better name.  What do you think?

Stebalien:
Builder sounds like a great name!

Mr0grog:
Sounds good to me :)

format.go
Stebalien: My point is that I really don't see why we can't just make `Prefix` implement `Format` (well, `Builder` now). Internally, it would just do *exactly* what the function above is doing and then call `Sum`. It just feels like adding this extra step isn't really necessary.

builder.go
```diff
@@ -0,0 +1,92 @@
...
+		mhLen = 0
+	}
+	return V1Builder{
+		Codec:   p.Codec,
+		HashFun: p.MhType,
+		HashLen: mhLen,
+	}
+}
+
+func (p Prefix) GetCodec() uint64 {
```
Stebalien: If `Prefix` is going to implement `Builder`, `PrefixToBuilder(p).Sum(...)` should be functionally equivalent to `p.Sum(...)`. My solution would be to just put the logic in `PrefixToBuilder` into `Prefix.Sum` but the alternative is to just *call* `PrefixToBuilder` in `Sum` (actually, that may just be the better solution).

builder.go
```diff
@@ -0,0 +1,92 @@
...
+	return p.Codec
+}
+
+func (p Prefix) WithCodec(c uint64) Builder {
```
Stebalien: Should comment that this will auto-upgrade to CIDv1 if necessary. Actually, I wonder if this should return a prefix or just return a builder (V1/V0).

kevina:
@Stebalien Prefix implements `Builder` right now.  My thoughts are that at some point we might want to depreciate this behavior and then the `PrefixToBuilder` function may be useful.  However, it if is causing you that much of a problem (or confusion) I can just remove `PrefixToBuilder`.

I do have one question though do you want to eventually depreciate Prefix.Sum?

builder.go
kevina: Yes that is intentional.  I don't see any reason why this should return a builder V1/V0.

kevina:
@Stebalien 

> However, I have some concerns about PrefixToBuilder(p).Sum not behaving the same way as p.Sum.

In what way?

kevina:
Okay.  I clean up the code to remove all the controversial parts so this should be ready for review.

Before we merge this it needs to be tested with `go-ipfs`.  I will update https://github.com/ipfs/go-ipfs/pull/5281 so that it complies with this p.r.

builder.go
Stebalien: It was just a thought. I don't really see any benefit one way or the other.

kevina:
> No. That is, I'd just treat Prefix as a valid builder.

I don't necessary agree with this, but that is something to revisit later.

Stebalien:
>> No. That is, I'd just treat Prefix as a valid builder.

> I don't necessary agree with this, but that is something to revisit later.

It may not make sense in all cases however, in those where it doesn't, I expect we'll want something fancier than a simple `PrefixToBuilder` function. That is, we'll want a smarter builder that's better about picking the hash function.

Basically, I agree that `Prefix` is not a good builder however, I don't see the builder produced by `PrefixToBuilder` as being any better (at least in it's current incarnation).

kevina:
Okay.  I completed the integration testing so this should be good to go.  Let's get #61 is also before we `gx publish`.

PR #55, opened 2018-07-11
Use value receiver for MarshalJSON so we can marshal values

phritz:
This is a problem for go-cbor-ipld, see https://github.com/filecoin-project/go-filecoin/issues/626 and https://github.com/filecoin-project/go-filecoin/issues/599.


Stebalien:
We really *shouldn't* be using MarshalJSON (we should have explicit transforms between the JSON format and the CBOR format)... but we're doing this anyways.

PR #56, opened 2018-07-11
release marshal bugfix

phritz:
Fix in https://github.com/ipfs/go-cid/pull/55 on which this change depends. I'll merge that one first when approved so this change is only the release.

In service of https://github.com/filecoin-project/go-filecoin/issues/626


Stebalien:
Pushed release to master (gx is special).

PR #60, opened 2018-07-26
Create a new Encode method that is like StringOfBase but never errors 

kevina:
The idea is that we can just replace calls to `String()` with `Encode(base)`.  For this to work we need a method that will return meaningful results for all CIDs and never errors.

Used by https://github.com/ipfs/go-ipfs/pull/5289.

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -338,6 +338,18 @@ func (c *Cid) StringOfBase(base mbase.Encoding) (string, error) {
 	}
 }
 
+// Format return the string representation of a Cid
+func (c *Cid) Format(base mbase.Encoder) string {
+	switch c.version {
+	case 0:
+		return c.hash.B58String()
+	case 1:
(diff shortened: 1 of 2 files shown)


cid.go
```diff
@@ -337,6 +337,19 @@ func (c *Cid) StringOfBase(base mbase.Encoding) (string, error) {
 	}
 }
 
+// Encode return the string representation of a Cid in a given base
+// when applicable
```
Stebalien: I'd definitely expand on the edge cases here. Other than that, I have no real objections if we end up needing this.

kevina:
@Stebalien this will end up being useful.  If we are going to push #71 though in the next couple of days I would like to get this in also.

Stebalien:
We should probably *also* deprecate `StringOfBase` eventually, but this is good for now.

PR #61, opened 2018-07-31
cid-fmt Enhancments

### LINKED ISSUE #41: Extract cid-fmt formatter out of main package so it can be used as a lib.

kevina:
Refactor and enhance cid-fmt and extract most of it from the main package so the functionality can be used as a library.

Closes #41.

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -213,34 +213,47 @@ func Parse(v interface{}) (*Cid, error) {
 // starting with "Qm" are considered CidV0 and treated directly
 // as B58-encoded multihashes.
 func Decode(v string) (*Cid, error) {
-	_, cid, err := DecodeV2(v)
-	return cid, err
-}
-
-// DecodeV2 is like Decide but also returns the Multibase encoding the
-// Cid was encoded in.  EXPERIMENTAL and interface may change at any time.
-func DecodeV2(v string) (mbase.Encoding, *Cid, error) {
 	if len(v) < 2 {
-		return 0, nil, ErrCidTooShort
+		return nil, ErrCidTooShort
 	}
 
 	if len(v) == 46 && v[:2] == "Qm" {
 		hash, err := mh.FromB58String(v)
 		if err != nil {
-			return 0, nil, err
+			return nil, err
 		}
 
-		return mbase.Base58BTC, NewCidV0(hash), nil
+		return NewCidV0(hash), nil
 	}
 
-	base, data, err := mbase.Decode(v)
+	_, data, err := mbase.Decode(v)
 	if err != nil {
-		return 0, nil, err
+		return nil, err
 	}
 
-	cid, err := Cast(data)
+	return Cast(data)
+}
 
-	return base, cid, err
+// Extract the encoding from a Cid.  If Decode on the same string did
+// not return an error neither will this function.
+func ExtractEncoding(v string) (mbase.Encoding, error) {
+	if len(v) < 2 {
+		return -1, ErrCidTooShort
+	}
+
+	if len(v) == 46 && v[:2] == "Qm" {
+		return mbase.Base58BTC, nil
+	}
+
+	encoding := mbase.Encoding(v[0])
+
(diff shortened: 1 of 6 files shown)


kevina:
Note: It adds a new `Format` function that conflicts with the name currently used in #53.  However I think Format is a better name for this than for the new interface in #53.

However, I am also thinking that this `Format` function doesn't really belong in the main `cid` package and should be moved to a separate package perhaps called `cidutil` and located in the `util/` directory of `go-cid` (so it will be imported with `import cidutil "github.com/ipfs/go-cid/util`.

Thoughts?

kevina:
Note #60 should go in first.

kevina:
Blocked #64 as it will make part of this code cleaner.

kevina:
With the new `Format` interface renames to `Builder` this can go in as is now.

kevina:
@Stebalien I would like to get this in about the same time we get #53 in.

format.go
```diff
@@ -0,0 +1,151 @@
...
+
+(1) For CID version 0 the multibase must be base58btc and no prefix is
+used.  For Cid version 1 the multibase prefix is included.
+`
+
+// Format formats a cid according to the format specificer as
+// documented in the FormatRef constant
+func Format(fmtStr string, base mb.Encoding, cid *Cid) (string, error) {
+	p := cid.Prefix()
+	out := new(bytes.Buffer)
```
Stebalien: No need to allocate here. We can just put this on the stack. It even has a small internal 64byte buffer to avoid allocations when formatting small things.

format.go
```diff
@@ -0,0 +1,151 @@
...
+	if !ok {
+		return fmt.Sprintf("codec?%d", num)
+	}
+	return name
+}
+
+func hashToString(num uint64) string {
+	name, ok := mh.Codes[num]
+	if !ok {
+		return fmt.Sprintf("hash?%d", num)
```
Stebalien: Personally, for performance, I'd rewrite these helpers to write directly to the buffer (e.g., `fmtHash(b bytes.Buffer, num uint64)`). However, we can always revisit this if it becomes a performance issue (it won't be for the moment).

Stebalien:
LGTM. I have a few small suggestions but I'd be happy merging as is.

format.go
```diff
@@ -0,0 +1,151 @@
...
+   %S cid string encoded in base %b without multibase prefix
+   %P cid prefix: %v-%c-%h-%L
+
+(1) For CID version 0 the multibase must be base58btc and no prefix is
+used.  For Cid version 1 the multibase prefix is included.
+`
+
+// Format formats a cid according to the format specificer as
+// documented in the FormatRef constant
+func Format(fmtStr string, base mb.Encoding, cid *Cid) (string, error) {
```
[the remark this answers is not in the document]
kevina: I think that is an overkill. :)  A better model might be the time package.  I could make Format a method, but I also think something like this should not be port of the core Cid functionally and possible, at a later date, moved to an external package.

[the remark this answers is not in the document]
Stebalien: Yeah... probably.

cid.go
```diff
@@ -234,6 +234,28 @@ func Decode(v string) (*Cid, error) {
 	return Cast(data)
 }
 
+// Extract the encoding from a Cid.  If Decode on the same string did
+// not return an error neither will this function.
+func ExtractEncoding(v string) (mbase.Encoding, error) {
```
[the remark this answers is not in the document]
kevina: I did that at first.  The ExtractEncoding is useful when you just want the encoding and it is fairly efferent and should not allocate any heap memory so the cost of the extra function call should be negligible.

format.go
kevina: In most cases a string from the look table is returned.  Sprintf is only for the rare corner case when an encoding was used that is not in the table.

format.go
kevina: Okay fixed.

PR #62, opened 2018-08-01
Feature: support storing a value along with the Cid

hsanjuan:
This adds support for storing ordered-pairs (a tuple of 2 elements, being the
first one the Cid) in the set.

A VisitPair function allows the user to decide, based on the second element
of the pairs, when to replace the existing element on the set (as opposed to the `Visit()` function which only considers if it existed previously).

---

@Stebalien this should allow shortcutting branches when exploring with depth-limits in go-ipfs, by storing the depth and providing a custom visit function, without the need for a special set type or another library. The set works and can be used as before.

### DIFF
--- a/set.go
+++ b/set.go
@@ -2,18 +2,25 @@ package cid
 
 // Set is a implementation of a set of Cids, that is, a structure
 // to which holds a single copy of every Cids that is added to it.
+// It also works with pairs formed by a Cid and a user provided arbitrary
+// value.
 type Set struct {
-	set map[string]struct{}
+	set map[string]interface{}
 }
 
 // NewSet initializes and returns a new Set.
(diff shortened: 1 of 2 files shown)


hsanjuan:
I'd say it's a set of pairs with a custom visit function. It allows doing just what you proposed by passing a custom function.

Do you want the new type in a new repo? or here? It will need to be used in several modules (like `*cid.Set` which will replace). And how would you name it?

hsanjuan:
Ah, I guess you actually mean to do things directly with a plain map where needed and replace *cid.Set.

Stebalien:
> Ah, I guess you actually mean to do things directly with a plain map where needed and replace *cid.Set.

I think that'll be more readable but I may be wrong.

PR #66, opened 2018-08-06
WIP: Store multibase prefix with CID

kevina:
The CID spec defines the multibase as part of the CID, yet we do not store it anywhere.

This fixes that,  It also creates a variable to change the default base for CIDv1.

This should make preserving the base of a CID a lot less painful for command like `ipfs resolve` and `ipfs pin add` or any command that takes a CID from the user.

Simplifier version of #64.

### DIFF
--- a/cid-fmt/main.go
+++ b/cid-fmt/main.go
@@ -137,8 +137,7 @@ func decode(v string) (mb.Encoding, *c.Cid, error) {
 	if err != nil {
 		return -1, cid, err
 	}
-	enc, _ := cid.Base()
-	return enc.Encoding(), cid, err
+	return cid.Base().Encoding(), cid, err
 }
 
 const ERR_STR = "!ERROR!"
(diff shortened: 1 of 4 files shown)


Stebalien:
So, CIDs always use a multibase *in text*. The spec has always been a bit unclear on this subject but I'd be strongly against any attempt to carry the multibase along with a CID. That mixes data with representation a bit too much for my tastes.

Ideally, on the command line, we'd:

1. Accept *any* multibase.
2. Return the multibase requested by the user.

kevina:
@Stebalien this doesn't need to go in right away.  Can we hold off on making a quick judgement.  I will right a full proposal (with several) alternatives on how to handle multibases later today.

kevina:
See https://github.com/ipfs/go-ipfs/issues/5349 solution 2 for the motivation for this change.

PR #69, opened 2018-08-17
Extract non-core functionality from go-cid into go-cidutil

### LINKED ISSUE #68: Extract non-core functionaly into new package
`go-cid` is used by any package that does anything IPFS or IPLD related.  we should probably move non-core functionally into a new package to avoid having to update so many packages when non-core functionality changes.

Code in the following packages should likely be moved:
  * `set.go` and tests
  * `format.go` and tests
  * Everything in `cid-fmt/`.

kevina:
Closes #68


kevina:
@Stebalien 

Before we merge and propitiate this let get https://github.com/ipfs/go-cidutil/pull/4.

The gx publish is pinned on one of my personal ipfs nodes and was required to properly test the extraction to `go-cidutil`.

PR #70, opened 2018-08-24
cid implementation roundup

warpfork:
It's been discussed in several issues and PRs already that we might want to explore various ways of implementing CIDs for maximum performance and ease-of-use because they show up extremely often.  Current CIDs are pointers, which generally speaking means you can't get one without a malloc; and also, they're not particularly well-suited for use in map keys.

This branch is to attempt to consolidate all the proposals so far  (https://github.com/ipfs/go-cid/issues/38, https://github.com/ipfs/go-cid/pull/64, https://github.com/ipfs/go-cid/pull/47, more?) -- and do so in a single branch which can be checked out and contains all the proposals at once, because this will make it easy to do benchmarks and compare all of the various ways we could implement this in one place (and also easier for humans to track what the latest of each proposal is, since they're all in one place).  (I ended up re-writing much of it because some of the existing work was a pain to cherry-pick apart, either because it was stale and had conflicts with latest master or because it mixed other topics like multibase, but this *does* borrow heavily from @dignifiedquire and @kevina 's code on the other branches.)

There are two/three major implementations here: one as struct (familiar), one as typedef of string, and one as an interface (implemented by the other two).

I composed a README.md with a summary of what I learned.  Some of it became obvious just from thrashing the code around; some of it is supported with benchmarks (also included in the diff).

There's no attempt here to change the implementation of multihash... however, it's likely that doing so could further shift some of the performance numbers quite a bit.  Skim the alloc counts in the benchmarks for impressions.

This whole hullabaloo is compartmentalized under a `_rsrch` directory... which means it won't typically be built or tested by e.g. `go test ./...` unless you're already in the `_rsrch` directory.  It also does not change any of the existing packages.  Therefore, it is completely safe to merge to master as sheer documentation of our research, even if we choose not to pursue any of these refactors at this time.

Hopefully this is good further food for thought and can be iterated on to get us closer to a committal choice about this stuff.  (Or at least merged as research, rather than abandoned to drift as an out-of-sync branch.  I dunno if you can tell, but those are my least favorite thing.)

### DIFF
--- a/_rsrch/cidiface/README.md
+++ b/_rsrch/cidiface/README.md
@@ -3,10 +3,7 @@ What golang Kinds work best to implement CIDs?
 
 There are many possible ways to implement CIDs.  This package explores them.
 
-- Option A: CIDs as a struct; multihash as bytes.
-- Option B: CIDs as a string.
-- Option C: CIDs as an interface with multiple implementors.
-- Option D: CIDs as a struct; multihash also as a struct or string.
+### criteria
 
 There's a couple different criteria to consider:
 
@@ -20,6 +17,34 @@ mapkeys > minalloc > anythingelse.
 representation that can work natively as a map key, we'll end up needing
 a `KeyRepr()` method which gives us something that does work as a map key,
 an that will almost certainly involve a malloc itself.)
+
+### options
+
+There are quite a few different ways to go:
+
+- Option A: CIDs as a struct; multihash as bytes.
+- Option B: CIDs as a string.
+- Option C: CIDs as an interface with multiple implementors.
+- Option D: CIDs as a struct; multihash also as a struct or string.
+- Option E: CIDs as a struct; content as strings plus offsets.
+
+The current approach on the master branch is Option A.
+
+Option D is distinctive from Option A because multihash as bytes transitively
+causes the CID struct to be non-comparible and thus not suitable for map keys
+as per https://golang.org/ref/spec#KeyType .  (It's also a bit more work to
+pursue Option D because it's just a bigger splash radius of change; but also,
+something we might also want to do soon, because we *do* also have these same
+map-key-usability concerns with multihash alone.)
+
+Option E is distinctive from Option D because Option E would always maintain
+the binary format of the cid internally, and so could yield it again without
+malloc, while still potentially having faster access to components than
+Option B since it wouldn't need to re-parse varints to access later fields.
+
+Option C is the avoid-choices choice, but note that interfaces are not free;
+since "minimize mallocs" is one of our major goals, we cannot use interfaces
+whimsically.
 
 
 Discoveries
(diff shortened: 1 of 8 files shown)


warpfork:
As a result of this, I'm mostly coming around to being a pretty vigorous advocate for `type Cid string`.

It might be nice to have a `type CidInterface interface{...}` at the same time... but its usage should be minimized.  In particular, I think `map[CidInterface]*` is a *really* bad idea and should be avoided at all costs.  (`func CanonicalizeCid(CidInterface) Cid` would be a fine thing to have, though.)

kevina:
> In particular, I think map[CidInterface]* is a really bad idea and should be avoided at all costs. 

I agree and it was never my intention to propose such usage.  That is what CidString is for.  It is an implementation of a Cid as a string for use in interfaces.

Stebalien:
Let's do it!

kevina:
@Stebalien 

> Let's do it!

What?

I would like a chance to review these benchmarks.  There where several things about @dignifiedquire that I did not like and changed when I implemented my version of it in #65. We do not have to go with an interface, but I would like my changes to the String representation to go in.

We should _really_ also change Multihash to be a string if a Cid is string, otherwise we we be creating a lot of needless alloc.  We should also consider enhancing the multibase interface to (in addition) provide functions that take a string so that we can avoid an alloc there, but I see that as a lower priority.

I would be happy to take charge of this.

kevina:
> Which changes?

There in #65.  I separate out the String stuff from just the interface stuff.

Stebalien:
Ah, carrying the multibase with the Cid? I've commented on that PR. I understand that it's *a* solution to the CID base issue but I'm far from convinced that it's the right one (or even a palatable one).

kevina:
> Ah, carrying the multibase with the Cid? I've commented on that PR. I understand that it's a solution to the CID base issue but I'm far from convinced that it's the right one (or even a palatable one).

I will create a new P.R. with _just_ the string stuff, stay tuned....

kevina:
@warpfork I would be interested in how much the extra allocations cost us when we have to get the multihash in the string implementation.  Also see #71 which is a continuation and optimization of #47.  In particular it avoids an alloc when calling Type() which could be significant depending on how the Cid is used.

warpfork:
Okay, so I'm gonna get ready to hit the merge button.  Reminder, from a purely workflow management perspective, the goal in this PR was to *make something mergeable* which does not need to be the full final solution -- note the absolute lack of change in the main package -- and can also be kept in history as research notes for why we choose what we do.  It shouldn't make merge conflicts with anything else in progress either.

P.S. yeah the uvarint impl speciated for strings in  #71... :+1: I bet that's going to be impactful.  I didn't write benchmarks here yet either for how much so, but there's no way evading the byte slice conversion there isn't gonna be helpful.

PR #71, opened 2018-08-25
continuation of use string instead of struct backing

### LINKED ISSUE #3: Suggestion to avoid exra layer of indirection by using a more compact representation of Cid
The conversion from plain `Key` to `Cid` introduced an extra layer of indirection that can likely be avoided.

Currently Cid takes up the width of 5*64 bits.  With it's current definition of:

```
type Cid struct {
    version uint64
    codec   uint64
    hash    mh.Multihash
}
```

May I suggest

```
type Cid struct {
    version uint32
    codec   uint32
    hash    mh.Multihash
}
```

As 32 bits should be more than enough for version and codec which I don't think will ever get very large (even 16 bits should be enough).  If combined with multiformats/go-multihash#29 the size will now be 3*64 bits, which is the same size an array slice.

As the same size of a slice it becomes practical to pass it around by value which avoids an extra layer of indirection.  Even though it is passed by value, the internals can still be kept hidden, as is done with many other go datatypes.

Just a suggestion.  It might make a difference when you have an array of 1000's of Cids (which I might in some of my maintenance commands for the filestore, ipfs/go-ipfs#2634).

### LINKED ISSUE #38: Suggestion: Change CID type so it can be passed around directly and not by pointer

kevina:
Rebased #47 and made some improvements.  Most of the changes where noted in my review of #47.

I tried to keep each change in it's own commit, so if any of my changes are not agreeable I can back them out.

Closes #3.  Closes #38.


kevina:
> This looks like a good way to do this. We should probably, eventually, do the same with Multihash but we can do that later.

The fact that the Cid type change from a pointer to a non-pointer type is going to make this change painful, I would vote to change multihash to a string at the same time and get it over with, but I am not of a strong option on this if you are eager to push this trough now.

cid.go
```diff
@@ -405,9 +400,8 @@ func (c *Cid) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
-	c.version = out.version
-	c.hash = out.hash
-	c.codec = out.codec
+	*c = Cid{out.string[:]}
```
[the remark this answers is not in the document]
kevina: Yes but i think you meant `*c = out`

kevina:
@Stebalien Probably, it might be easier to do it at the same time though.

cid.go
```diff
@@ -132,36 +132,40 @@ var CodecToStr = map[uint64]string{
...
-	codec   uint64
-	hash    mh.Multihash
-}
+// Byte layout: [version, codec, multihash]
+// - version uvarint
+// - codec uvarint
+// - hash mh.Multihash
+type Cid struct{ str string }
+
+var Nil = Cid{}
```
warpfork: How about calling this `Zero`?

"Zero" would more closely match https://golang.org/ref/spec#The_zero_value

kevina: `Nil` is a more intuitive name to me.  `Zero` implies a numeric value to me.

Stebalien:
@kevina up to you.

kevina:
> @kevina up to you.

Once we switch to using Multihashes in the blockstore, the `cid.Hash()` method is going to be called a lot more. To prevent performance regressions it would probably be a good idea if that method doesn't alloc.

Stebalien:
Go ahead.

kevina:
> Go ahead.

Done. See https://github.com/multiformats/go-multihash/pull/82

Stebalien:
> Probably, it might be easier to do it at the same time though.

Given that this isn't looking to be the case, let's go ahead with this and then try tackling the multihash issue.

kevina:
> Given that this isn't looking to be the case, let's go ahead with this and then try tackling the multihash issue.

I left an additional option on the p.r. that might make this easier.

Please let me know what you intended time frame for landing this is.  My current plan was to add the `--cid-base` to go-ipfs and then swing back around to merging this p.r. and fixing any broken code along the way.

Stebalien:
I'd like to get this done sooner or later. It's blocking https://github.com/ipfs/go-ipld-cbor/pull/30.

cid.go
```diff
@@ -132,36 +132,49 @@ var CodecToStr = map[uint64]string{
...
-	hash    mh.Multihash
+// Byte layout: [version, codec, multihash]
+// - version uvarint
+// - codec uvarint
+// - hash mh.Multihash
+type Cid struct{ str string }
+
+// Nil can be used to represent a nil Cid, using Cid{} directly is
+// also acceptable.
+var Nil = Cid{}
```
Stebalien: @warpfork this is where you'd object to `Nil`.

You are correct in that go would call this the "Zero" value.

Arguments against Zero:

* To me, "Zero" implies a number.
* `Zero` could mean the CID for the number 0 (`z7xsiH4JqWQudech83`). I expect that users will define "constant" CIDs all the time although this one might be a stretch.
* We've historically used `nil` for "empty CIDs".
* We have a CIDv0 ("version zero").
* `Zero` "looks funny" (entirely subjective) in the return position.

```
func Foo() (Cid, error) {
    return Nil, fmt.Errorf("bad stuff")
}
func Bar() (Cid, error) {
    return Zero, fmt.Errorf("bad stuff")
}
```

Stebalien (commented):
There are probably some optimizations we can toy with later but this looks good for now (from my end).

cid.go
```diff
@@ -132,36 +132,49 @@ var CodecToStr = map[uint64]string{
...
 
 // Cid represents a self-describing content adressed
 // identifier. It is formed by a Version, a Codec (which indicates
 // a multicodec-packed content type) and a Multihash.
-type Cid struct {
-	version uint64
-	codec   uint64
-	hash    mh.Multihash
+// Byte layout: [version, codec, multihash]
+// - version uvarint
```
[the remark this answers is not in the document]
kevina: Your right.  This is leftover from the original implementation that encoded a CIDv0 as if it where CIDv1.

kevina:
> There are probably some optimizations we can toy with later but this looks good for now (from my end)

Exactly what did you have in mind for this?

Stebalien:
> Exactly what did you have in mind for this?

* We can assume that the version is one byte (and check against the constant 0x1).
* We can probably rearrange some of the logic related to `Prefix()` to avoid parsing varints multiple times.
* etc...

But let's handle that later. The key part here is changing the interfaces.

kevina:
> We can assume that the version is one byte (and check against the constant 0x1).

Actually I already did this, but it was part of the commit the used the new Multihash.

> We can probably rearrange some of the logic related to Prefix() to avoid parsing varints multiple times.

Ditto. I think.

cid.go
Stebalien: You are correct in that go would call this the "Zero" value. However, I really don't like it.

Arguments against Zero:

* To me, "Zero" implies a number.
* `Zero` could mean the CID for the number 0 (`z7xsiH4JqWQudech83`). I expect that users will define "constant" CIDs all the time although this one might be a stretch.
* We've historically used `nil` for "empty CIDs".
* We have a CIDv0 ("version zero").
* `Zero` "looks funny" (entirely subjective) in the return position.

```
func Foo() (Cid, error) {
    return Nil, fmt.Errorf("bad stuff")
}
func Bar() (Cid, error) {
    return Zero, fmt.Errorf("bad stuff")
}
```

Stebalien:
@kevina want to go ahead and do the refactor?

Stebalien:
> I'm confused?

No no, I'm talking about the whole cid type refactor (that is, fixing all the dependent packages). We can do any additional optimizations to this package later.

kevina:
Sure, not sure I will get to it tonight though.

set.go
```diff
@@ -55,9 +54,9 @@ func (s *Set) Visit(c *Cid) bool {
 
 // ForEach allows to run a custom function on each
 // Cid in the set.
-func (s *Set) ForEach(f func(c *Cid) error) error {
+func (s *Set) ForEach(f func(c Cid) error) error {
 	for cs := range s.set {
-		c, _ := Cast([]byte(cs))
+		c, _ := Cast(cs.Bytes())
```
Stebalien: We can just use the cid as-is (this is a left-over from when we needed to reconstruct it).

Stebalien (dismissed):
One issue, but otherwise LGTM.

Stebalien:
Note: Nobody but @kevina should merge this!

kevina:
@Stebalien okay fixed.  As I said in IRC I will first merge and bubble this one up and then do the larger go-multihash change shortly after.  I bubbled up go-cid before so I am sure I can handle it.  Go-multihash, maybe not but I will try.

cid.go
warpfork: Some other options just to toss out:

- `cid.Invalid` -- this shows up in a lot of other idiomatic Go code... but mostly in enum situations, so maybe less appropriate here.
- `cid.None` -- avoids all the numeric connotations, less "looks funny" in returns I think?

kevina: I still like `Nil` better...

kevina:
@Stebalien I changed the `IsNil` method to `Defined` when pushing this change through I found:
```
if c.Defined() {...}
```
easier to read then
```
if !c.IsNil() {...}
```
If you don't agree I can change it back.

cid.go
warpfork: I could go for `Undef` :+1:

Stebalien: :-1: on undef and invalid. The value *isn't* undefined and isn't the *only* invalid value. I'd prefer `None` (although I still prefer `Nil`).

Stebalien: Actually, seeing `cid.Defined()`, I'm onboard with this (if we *return* `Undef` and *check* with `cid.Defined()`).

Stebalien:
@kevina I agree. But if we do this we should change the sentinal value to `Undef` or `Undefined`.

cid.go
kevina: Okay I am going with `Undef`.

cid.go
```diff
@@ -132,36 +132,54 @@ var CodecToStr = map[uint64]string{
...
+	if cn != hashlen {
+		panic("copy hash length is inconsistent")
 	}
+
+	return Cid{string(buf[:n+hashlen])}
 }
 
 // Cid represents a self-describing content adressed
 // identifier. It is formed by a Version, a Codec (which indicates
 // a multicodec-packed content type) and a Multihash.
```
lanzafame: @kevina Could you please add a high level description of how the `version`, `codec`, and `hash` are packed into a string, similar to how the [multihash docs](https://multiformats.io/multihash/#the-multihash-format) do, to the godoc.

kevina: We decided against that.  The format varies according to the CID type (see https://github.com/ipfs/go-cid/pull/71#discussion_r213883739).  The layout is according to the cid spec. see https://github.com/ipld/cid.

lanzafame: Fair, thoughts on adding a link to the cid spec to the comment then? The location of various specs isn't the easiest thing to deduce, i.e. the cid spec is under the ipld org whereas the implementation is under the ipfs org. Just a suggestion, not a blocker.

kevina: There entire documentation could use cleaning up. The format is mentioned in multiple places and the link to the spec is in the documentation for the package.  https://godoc.org/github.com/ipfs/go-cid.

cid.go
```diff
@@ -133,6 +133,15 @@ var CodecToStr = map[uint64]string{
...
+	// Need to make sure hash is valid for CidV0 otherwise we will
+	// incorrectly detect it as CidV1 in the Version() method
+	dec, err := mh.Decode(mhash)
+	if err != nil {
+		panic(err)
+	}
+	if dec.Code != mh.SHA2_256 || dec.Length != 32 {
+		panic("invalid hash for cidv0")
+	}
 	return Cid{string(mhash)}
```
kevina: @Stebalien I panic here in order to avoid introducing another API change.  However, the patch to strict CIDs  (#40) does change the return type to `(Cid, error)` for both NewCidV0 and NewCidV1.  Should we just change them and get it now and get it over with?

I don't have strong felling either way?

kevina: Okay, the only reason I suggested doing it now is to avoid another API change in the near future.  But we can keep this simple and change the return type later.

Stebalien: I just don't want to drag this out.

Stebalien:
@kevina are we going to need to bubble that multibase update? If so, we should do that *first*. The list of changes we'll need to make for that will be much larger.

kevina:
@Stebalien actually it won't I checked.

Using my new tool:
```
go-ipfs$ gx-update-helper preview go-multibase
github.com/multiformats/go-multibase

github.com/ipfs/go-cid :: go-multibase

github.com/ipfs/go-block-format :: go-cid
github.com/ipfs/go-cidutil :: go-cid
github.com/ipfs/go-ipfs-ds-help :: go-cid
github.com/libp2p/go-libp2p-routing :: go-cid
github.com/ipfs/go-verifcid :: go-cid

github.com/ipfs/go-ipfs-blocksutil :: go-block-format
github.com/ipfs/go-ipfs-chunker :: go-block-format
github.com/ipfs/go-ipfs-exchange-interface :: go-block-format
github.com/ipfs/go-ipld-format :: go-block-format
github.com/libp2p/go-libp2p-kad-dht :: go-libp2p-routing
github.com/libp2p/go-libp2p-routing-helpers :: go-libp2p-routing
github.com/ipfs/go-ipfs-blockstore :: go-block-format go-ipfs-ds-help
github.com/ipfs/go-ipfs-routing :: go-ipfs-ds-help go-libp2p-routing

github.com/ipfs/go-ipfs-posinfo :: go-ipld-format
github.com/ipfs/go-ipld-cbor :: go-ipld-format
github.com/ipfs/go-ipld-git :: go-ipld-format
github.com/libp2p/go-libp2p-pubsub-router :: go-libp2p-routing-helpers
github.com/ipfs/go-ipfs-exchange-offline :: go-ipfs-blockstore go-ipfs-blocksutil go-ipfs-exchange-interface
github.com/ipfs/go-bitswap :: go-ipfs-blockstore go-ipfs-blocksutil go-ipfs-exchange-interface go-ipfs-routing

github.com/ipfs/go-blockservice :: go-bitswap go-ipfs-exchange-offline

github.com/ipfs/go-merkledag :: go-blockservice

github.com/ipfs/go-path :: go-merkledag
github.com/ipfs/go-unixfs :: go-merkledag

github.com/ipfs/go-mfs :: go-path go-unixfs

github.com/ipfs/go-ipfs :: go-mfs
```
Which is saying anything that depends on `go-multibase` *also* depends on `go-cid`.

Stebalien:
Hm. Yeah, you're right. But whatever we do, let's really try to avoid tacking on random additions at the last minute.

PR #72, opened 2018-08-27
cid implementation variations++

warpfork:
Enumerating another option -- "Option F" -- for `type struct{string}`.

I originally got nerdsniped into looking at this because of [a comment elsewhere](https://github.com/ipfs/go-path/blob/a8774b68353d38d492a9a918bfa552c4f52f7ff1/path.go#L30-L31); it's also what the tip of #71 is doing.  Definitely deserves to be uniquely identified as an option.

I didn't bother to write a whole new implementation in this package of this thing, because it would be *almost* identical to the string typedef impl... but the exploratory inspection code I came up with in the playground is easy to keep in the lab notebook, so here it is.

### DIFF
--- a/_rsrch/cidiface/README.md
+++ b/_rsrch/cidiface/README.md
@@ -27,6 +27,7 @@ There are quite a few different ways to go:
 - Option C: CIDs as an interface with multiple implementors.
 - Option D: CIDs as a struct; multihash also as a struct or string.
 - Option E: CIDs as a struct; content as strings plus offsets.
+- Option F: CIDs as a struct wrapping only a string.
 
 The current approach on the master branch is Option A.
 
@@ -41,6 +42,10 @@ Option E is distinctive from Option D because Option E would always maintain
 the binary format of the cid internally, and so could yield it again without
 malloc, while still potentially having faster access to components than
 Option B since it wouldn't need to re-parse varints to access later fields.
+
+Option F is actually a varation of Option B; it's distinctive from the other
+struct options because it is proposing *literally* `struct{ x string }` as
+the type, with no additional fields for components nor offsets.
 
 Option C is the avoid-choices choice, but note that interfaces are not free;
 since "minimize mallocs" is one of our major goals, we cannot use interfaces
@@ -72,6 +77,7 @@ When using `*Cid`, the nil value is a clear sentinel for 'invalid';
 when using `type Cid string`, the zero value is a clear sentinel;
 when using `type Cid struct` per Option A or D... the only valid check is
 for a nil multihash field, since version=0 and codec=0 are both valid values.
+When using `type Cid struct{string}` per Option F, zero is a clear sentinel.
 
 ### usability as a map key is important
 
@@ -82,6 +88,7 @@ We already covered this in the criteria section, but for clarity:
 - Option C: ~ (caveats, and depends on concrete impl)
 - Option D: ✔
 - Option E: ✔
+- Option F: ✔
 
 ### living without offsets requires parsing
 
@@ -106,7 +113,7 @@ How much this overhead is significant is hard to say from microbenchmarking;
 it depends largely on usage patterns. If these traversals are a significant
 timesink, it would be an argument for Option D/E.
 If these traversals are *not* a significant timesink, we might be wiser
-to keep to Option B, because keeping a struct full of offsets will add several
+to keep to Option B/F, because keeping a struct full of offsets will add several
 words of memory usage per CID, and we keep a *lot* of CIDs.
 
 ### interfaces cause boxing which is a significant performance cost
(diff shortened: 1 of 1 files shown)


_rsrch/cidiface/README.md
```diff
@@ -129,6 +136,23 @@ if a situation is at the scale where it's become important to mind whether
...
+precisely the same amount of memory.
+
+This is interesting because it means we can choose between either
+type definition with no significant overhead anywhere we use it:
+thus, we can choose freely between Option B and Option F based on which
+we feel is more pleasant to work with.
+
+Option F (a struct wrapper) means we can prevent casting into our Cid type.
+Option B (typedef string) can be declared a `const`.
+Are there any other concerns that would separate the two choices?
```
Stebalien: The struct wrapper allows us to add additional metadata/offsets later if we want.

warpfork: I was going to speculate about if one would be easier/faster for refmt to munge than the other, because struct-containing-string seems like it going to require at least a few more reflection calls...

But on second thought that probably doesn't matter at all; CIDs are going to be using [transform funcs](https://godoc.org/github.com/polydawn/refmt/obj/atlas#UnmarshalTransformFunc) either way because the serial type is either bytes (which is a different `reflect.Kind` than string, so needs transform) or base-$n$ encoded string (which... needs to be de-base-$n$'d, so that's a transform).  So it's the same.

*Are* there any other things that would make B/F distinct?  :thinking:

Stebalien: People would still compare against "" and cast to a string/bytes.

warpfork: The former should be rejected in code review based on the existence of `cid.Zero` and introduced universally when we do the refactor; the latter would be a *bug*.

If we're seriously worried about these, then we should 100% hands-down prefer Option F over B, because more than precisely zero instances of casting to Cid from outside the cid package would strike me as acceptable code quality.

Stebalien: > The former should be rejected in code review based on the existence of cid.Zero and introduced universally when we do the refactor; the latter would be a bug.

That's not an assumption we can make, generally. Users (not in our projects) *will* cast. It's really the *primary* reason to use a type alias over a struct.

kevina: After thinking about this, the face that a struct value can not be declared a const (another misfeature of go in my option) might have a performance impact when comparing to the zero value (`cid.Nil`).  Since the zero value is not a const the go compiler can't optimize on that fact and may have to do a load from memory.

kevina: > But I doubt it's a serious performance thing. 

Probably not.  But if we commit to making it a struct we could just use cid.Cid{}.  It is just a clear and avoids the problem altogether.

Stebalien: I don't think it'll make a difference, as long as we never modify it.

kevina: For the choice of the constant I am going with `Nil`.  My second choice would be `Empty`, and `Zero` the third.  `Nil` to me is neutral and also implies an invalid value, of which an empty Cid{} struct is (it can't be used for anything).  `Empty` to me implies a valid but empty container, but I am okay with it if `Nil` is causing that much of a problem.  `Zero` to me implies a numeric value, of which a Cid is not.  I understand the notion of a zero-value in compiler terminology, but `Zero` still strikes me as an odd choice, `ZeroValue` may be a _bit_ better, but if `Nil` is no good I still rather go with `Empty`.

I didn't realize that using struct initializers in the middle of equality was so verbose.  However, it was never my intent to use them in equality.  Instead I provided an `IsNil` method as c.IsNil() is nicer than c == cid.Nil or c == (cid.Cid{}).  However elected to still provide `var Nil = Cid{}` so it can be used if desired.

warpfork: If we're going to make speculations about whether comparison to a var is going to be a performance issue, then shouldn't proposing an equality *method* have at least as much burden to prove itself a non-issue before we consider that a confirmed design choice?

I disagree completely about naming a variable of non-pointer type "Nil".  That is extremely unidiomatic Go because it conflicts with the language's own definitions of the term, and in discussion elsewhere I've already linked the part of the language spec which defines zero versus nil.  But this PR is probably not the topical place to relitigate this naming, either, seeing as how the term is not used anywhere in the diff body we're discussing.

warpfork:
I think this has generated some good discussion, and we could merge it now.

In review: The diff body improves the README by enumerating a new option; it mentions the casting and const issues; and it provides some evidence of memory layout impact.  Thus, it's strictly better documentation and more information than before.

Since we don't seem to be coming up with any more distinctions between B/F, and just thinking out loud about how we balance them, I think we can move on.  As with the prior PR, this is all contained in a directory which is research content only and affects no APIs, so there's no splash radius to consider before merging.

_rsrch/cidiface/README.md
Stebalien: I'm strongly in favor of `Nil` as a constant. While `Zero` is technically correct, it's confusing.

However, as a method, I'd be fine with `IsEmpty` (or `IsZero` but I'd rather not).

As for the method, golang is pretty good about inlining one-line methods so it should be fine. After doing a bit of decompiling, it looks like we should go with:

```func
func (c Cid) IsNil() bool {
    return c.string == ""
}
```

`c == (Cid{})` seems to cost us an extra xor zeroing something out. `c == Nil` costs us a mov.

See: https://go.godbolt.org/z/CHyhZl

whyrusleeping: @Stebalien the `IsNilString` one looks wrong. The compiler is getting really smart there and looks like its comparing the empty string to itself, and calling that the output.

Stebalien: (discussed on slack)

The assembly was correct. The generated instruction was `testq AX, AX` which tests `AX & AX` to see if it's non-nil.

PR #73, opened 2018-08-29
Update to use the new string representation of a Multihash.

kevina:
Factored out of #47


kevina:
@Stebalien can you have a look at this.  This change needs to go in if we are going to update to the string representation of a multihash.  It also has a few other optimizations in.

kevina:
Note: This is based on #71.  Please approve but don't merge.  It has a gx-depenecy on a non-pinned hash that I will get ride of once https://github.com/multiformats/go-multihash/pull/84 is merged.

ribasushi:
@Stebalien is this still relevant? Seems everything has already migrated to strings...

Stebalien:
The goal was to _also_ migrate multihash.

PR #81, opened 2019-02-19
Let Cid implement Binary[Un]Marshaler and Text[Un]Marshaler interfaces.

hsanjuan:
This makes Cid implement https://golang.org/pkg/encoding/#BinaryMarshaler
which is used by go-codec to decide if things know how to serialize themselves
(currently we need do manual wrapping for anything containing a CID).

Since I was at it, I did the TextMarshaling one too.

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -21,6 +21,7 @@ package cid
 
 import (
 	"bytes"
+	"encoding"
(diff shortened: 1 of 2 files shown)


hsanjuan:
Thanks

PR #84, opened 2019-02-21
fix inline CIDs generated by Prefix.Sum

Stebalien:


Stebalien:
Yes, my code is bestest code.

PR #85, opened 2019-05-06
default cidv1 to base32

Stebalien:
https://github.com/ipfs/go-ipfs/issues/6220

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ This will make sure that dependencies are rewritten to known working versions.
 
 ```go
 // Create a cid from a marshaled string
-c, err := cid.Decode("zdvgqEMYmNeH5fKciougvQcfzMcNjF3Z1tPouJ8C7pc3pe63k")
+c, err := cid.Decode("bafzbeigai3eoy2ccc7ybwjfz5r3rdxqrinwi4rwytly24tdbh6yk7zslrm")
 if err != nil {...}
 
 fmt.Println("Got CID: ", c)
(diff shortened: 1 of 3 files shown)


Stebalien:
@whyrusleeping or @anorth,

We'd like to make go-cid return base32 CIDv1 CIDs by default instead of base58 (CIDv0 CIDs will continue to use base58). The argument is that:

1. Not many users are using CIDv1 yet so this shouldn't be to obvious of a change.
2. We'd like to switch to base32 by default eventually as it's case-insensitive. There's no reason not to switch now.

Proposal here: https://github.com/ipfs/go-ipfs/issues/6220

---

As far as I can tell, the go-filecoin project doesn't test against any hard-coded base32 CIDv1 CIDs so this shouldn't have any impact (other than maybe changing the CLI output in some cases). However, I figured you'd like to be aware of this change and have a chance to raise objections.

If you _do_ have any repos with a bunch of base58 CIDv1 CIDs that you'd like to convert, the following perl script works pretty well (but double check the results, it has false positives):

```
perl -i -e 'sub base32 { my ($cid) = @_; my $res=`ipfs cid base32 $cid`; chomp $res; return $res; }' -pe 's/\b(z[a-zA-Z0-9]{10,})/"@{[base32($1)]}"/ge' **/*
```

whyrusleeping:
SGTM, ship it!

anorth:
No objections, thanks for the heads-up!

Stebalien:
@lidel I fixed the readme but couldn't find the CIDs you're referring to in the tests.

hsanjuan:
Heads up: breaking change tagged with patch release. People may rely on what cid strings look like (for me it just breaks sharness tests).

Stebalien:
Apologies, I should have bumped the minor.

PR #94, opened 2019-11-08
docs: example on how to create multiple V0 CIDs

daviddias:
I add to figure it out, so leaving it here for others :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,27 @@ if !c.Equals(other) {
 
 ```
 
+#### Create random V0 CIDs to play with and for testing
+
+```
+import (
+	"fmt"
+
+	"github.com/ipfs/go-cid"
+	ipfsUtil "github.com/ipfs/go-ipfs-util"
+)
+
+func main () {
+	var randomCids []cid.Cid
+
+	for i := 0; i < 10; i++ {
+		v := fmt.Sprintf("%d -- some string to spicy things up", i)
+		mhv := ipfsUtil.Hash([]byte(v))
+		randomCids = append(randomCids, cid.NewCidV0(mhv))
+	}
+}
+```
+
 ## Contribute
 
 PRs are welcome!


README.md
```diff
@@ -117,6 +117,27 @@ if !c.Equals(other) {
...
 ```
 
+#### Create random V0 CIDs to play with and for testing
+
+```
+import (
+	"fmt"
+
+	"github.com/ipfs/go-cid"
+	ipfsUtil "github.com/ipfs/go-ipfs-util"
```
Stebalien: So... the best way to do this is currently:

```go
(cid.V0Builder{}).Sum([]byte(v))
```

Or, more accurately for "raw data":

```go
(cid.V1Builder{Codec: cid.Raw, MhType: mh.SHA_256}).Sum([]byte(v))
```
(where `mh` is `github.com/multiformats/go-multihash`).

I'd like to eventually kill off `go-ipfs-util`.

Stebalien:
Thanks!

README.md
daviddias: Just experimented with these and although they work, they felt really "ninja". I believe that people will get more value from learning how to create multiple cids from a multihash operation, then _hacking_ a cid.V0 object out a straight raw byte buffer.

I mean, as a reader, decomposing:

```
(cid.V0Builder{}).Sum([]byte(v))
```

and 

```
(cid.V1Builder{Codec: cid.Raw, MhType: mh.SHA_256}).Sum([]byte(v))
```

Is hard. Questions that automatically pop up are:
- Why `Sum`?
- Should I trust that this CID builder will multihash the buffer? The V1 suggests yes, but the V0 suggests that it will just treat that byte array as a multihash, which is not.

Stebalien: `Sum` is "checksum". It's the term that go uses in the standard library: https://godoc.org/hash#Hash

I agree it's a bit funky but `ipfsUtil.Hash` always creates a V0 CID, regardless of what we're hashing.

PR #95, opened 2019-11-11
Add a function to read a cid from an array of bytes

### LINKED ISSUE #93: Support for reading concatenated CIDs
CIDs carry a prefix including the length of the multihash payload, so it should be possible to read concatenated CIDs from a stream or buffer without the need for additional metadata or delimiters. Unfortunately the methods available don't expose quite enough information to achieve this.

Some ways this could be fixed:
- have `Cast` return the number of bytes actually used from the argument slice
- add a new `Read(io.Reader) Cid` that leaves the input reader positioned at the first byte after the CID consumed

whyrusleeping:
closes #93

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -290,36 +290,16 @@ func uvError(read int) error {
 // Please use decode when parsing a regular Cid string, as Cast does not
 // expect multibase-encoded data. Cast accepts the output of Cid.Bytes().
 func Cast(data []byte) (Cid, error) {
-	if len(data) == 34 && data[0] == 18 && data[1] == 32 {
-		h, err := mh.Cast(data)
-		if err != nil {
-			return Undef, err
-		}
-
-		return NewCidV0(h), nil
-	}
-
-	vers, n := binary.Uvarint(data)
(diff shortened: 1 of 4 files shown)


cid.go
```diff
@@ -601,3 +581,42 @@ func PrefixFromBytes(buf []byte) (Prefix, error) {
...
+		return 0, Undef, fmt.Errorf("expected 1 as the cid version number, got: %d", vers)
+	}
+
+	_, cn := binary.Uvarint(data[n:])
+	if err := uvError(cn); err != nil {
+		return 0, Undef, err
+	}
+
+	mhr := mh.NewReader(bytes.NewReader(data[n+cn:]))
+	h, err := mhr.ReadMultihash()
```
anorth: Is this any better than `mh.Cast(data[n+cn:])`?

whyrusleeping: mh.Cast didnt work at all with trailing bytes, the Reader did, but allocated a lot more. I implemented a method in multihash to do just what we want here, and am using it now.

PR #99, opened 2020-02-04
fix: enforce minimal encoding

Stebalien:
Note: this removes two errors that don't appear to be used by anyone.


cid.go
```diff
@@ -173,9 +164,9 @@ func NewCidV0(mhash mh.Multihash) Cid {
 func NewCidV1(codecType uint64, mhash mh.Multihash) Cid {
 	hashlen := len(mhash)
 	// two 8 bytes (max) numbers plus hash
-	buf := make([]byte, 2*binary.MaxVarintLen64+hashlen)
-	n := binary.PutUvarint(buf, 1)
-	n += binary.PutUvarint(buf[n:], codecType)
+	buf := make([]byte, 1+varint.UvarintSize(codecType)+hashlen)
```
Kubuxu: I was going to say, "let's just keep it a static allocation as it gives compiler more freedom" but it is already `+hashlen` so it doesn't matter.

Stebalien:
Thanks @Kubuxu!

PR #102, opened 2020-05-01
add a couple useful methods

whyrusleeping:


cid.go
```diff
@@ -408,6 +409,17 @@ func (c Cid) Bytes() []byte {
...
 }
 
+// ByteLen returns len(c.Bytes()) without an allocation
+func (c Cid) ByteLen() int {
+	return len(c.str)
+}
+
+// WriteTo writes the cids bytes to the given writer
+func (c Cid) WriteTo(w io.Writer) error {
+	_, err := io.WriteString(w, c.str)
```
willscott, line 419: worth checking if this has fully written the string?

whyrusleeping: actually, probably better just to satisfy the WriterTo interface and return the write size

willscott: this won't work with the IO methods unless it has a loop as i'm reading the interface contract.
It should either go until there's an error, or until it's written the full string - right?

whyrusleeping: mmm, i think youre right there.... Maybe i'll pick a different function name. I don't particularly care for loops here

PR #104, opened 2020-05-12
feat: add Filecoin multicodecs

rvagg:
* fil-commitment-unsealed
* fil-commitment-sealed

Ref: https://github.com/multiformats/multicodec/pull/161
Ref: https://github.com/multiformats/multicodec/pull/172


rvagg:
I can't get the Travis Branch run to happen, I force pushed a commit with an updated timestamp and it's still stuck on "Expected - Waiting for status to be reported". Travis PR runs fine though.

Stebalien:
:man_shrugging:

rvagg:
@Stebalien I think this is good to land, could you merge it please?

rvagg:
The short answer is "partly". But we also purposefully didn't tag them as "ipld", or "serialization" in the multicodec table because (a) it's not as clear that these can be a clean & bi-directional serialization format and (b) it's not even clear that you'd ever want to use them in this way. We also still haven't received a clear answer regarding _why_ these things need to be CIDs but we're going forward in good faith that there is (and that grokking the full "why" of any of these things is probably beyond the scope of multicodecs in general if multicodecs are to continue to scale). _A pointer to a content-addressed thing of a particular, discernible type whose address is composed in a particular way_ - is partly (but not wholly) how I think it's being conceptualised.

Deep context can be found in this thread if anyone's interested in diving head-first: https://github.com/multiformats/multicodec/pull/172

PR #107, opened 2020-05-25
Chore/base36 support

ribasushi:
Bumps go-multibase and cleans up gx


package.json
```diff
@@ -3,28 +3,8 @@
   "bugs": {
```
Stebalien, line 3: We should just remove this file.

ribasushi:
@lidel it actually supports both/mixed case:
sha2-256-256-1e7bbdc0fa65843fb589401c4ef41cda75020b9a0075247850e222b9ec6f3463  ;)

PR #109, opened 2020-07-23
feat: optimize cid.Prefix

Stebalien:
We call this frequently and having to allocate here is really unfortunate.

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -518,12 +518,29 @@ func (c Cid) Loggable() map[string]interface{} {
 
 // Prefix builds and returns a Prefix out of a Cid.
 func (c Cid) Prefix() Prefix {
-	dec, _ := mh.Decode(c.Hash()) // assuming we got a valid multiaddr, this will not error
+	if c.Version() == 0 {
+		return Prefix{
+			MhType:   mh.SHA2_256,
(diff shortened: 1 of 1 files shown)


cid.go
```diff
@@ -518,12 +518,29 @@ func (c Cid) Loggable() map[string]interface{} {
...
+	offset := 0
+	version, n, _ := uvarint(c.str[offset:])
+	offset += n
+	codec, n, _ := uvarint(c.str[offset:])
+	offset += n
+	mhtype, n, _ := uvarint(c.str[offset:])
+	offset += n
+	mhlen, _, _ := uvarint(c.str[offset:])
+
 	return Prefix{
```
willscott, line 539: could allocate this struct before the uvarint's above, and then read version / type / codec directly into the member fields

whyrusleeping: eh, it won't save us anything. Its all in registers at the end of the day

Kubuxu:
@Stebalien are we sure that varint in the `c.str` is well structured? Otherwise `uvarint` witll return negative n.

Kubuxu:
Ahh, doc is bad. `uvarint` doesn't return negative n anymore.

PR #112, opened 2020-10-01
Add the dagjose multiformat

alexjg:


rvagg (approved):
`DagCBOR` sets an unfortunate precedent here for capitalistaion, it really should be `DagCbor` and therefore `DagJose` so that they're all predictable and potentially replaceable in the future with some automatic process that translates the table. But I guess it is what it is now so :shrug:.

/cc @mvdan for review too

mvdan (approved):
The Go SGTM. Why do we need to go from string to integer and back to string, by the way?

alexjg:
I guess that's so that client code can generically go from codec name to the actual multicodec value? E.g this code in `go-ipfs`:

```go
		if codecStr != "" {
			codec, ok := cid.Codecs[codecStr]
			if !ok {
				return fmt.Errorf("unknown IPLD codec: %s", codecStr)
			}
			opts.newCodec = codec
		} // otherwise, leave it as 0 (not a valid IPLD codec)
```

alexjg:
Hey folks, is there anything more that needs doing on this? Would be great to get it merged.

rvagg:
I've just realised I have merge access on this repo so could get this merged. There's something wrong with the codecov report though and I can't re-run it. Would you mind bumping the commit so it'll re-run please? The repo rules stop me overriding it unfortunately.

alexjg:
Hmm, I've tried bumping the commit and it looks like it's still causing problems. I can't really figure out why there are coverage changes, the codecov output doesn't seem particularly enlightening.

rvagg:
Your commit _accidentally_ slipped in via #115 (I must have been messing around with trying to get this in from the cmdline and had it on my master so it slipped in via a rebase). But that's OK because I was trying to get it in anyway! Done now.

PR #114, opened 2020-10-12
fix: make tests run with Go 1.5

rvagg:
> ./cid_test.go:451:52: conversion from uint64 to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)

Review please, @mvdan?


mvdan (approved):
LGTM. Please amend the commit message to say 1.15 instead, which is I think what you meant :)

rvagg:
in via #115

PR #116, opened 2020-12-01
fix: improved error message on broken CIDv0

lidel:
This aims to close DX/UX issue described in https://github.com/ipfs/go-ipfs/issues/7792

### Before

`invalid path "/ipfs/qmbwqxbekc3p8tqskc98xmwnzrzdtrlmimpl8wbutgsmnr/": invalid CID: selected encoding not supported`

### After

`invalid path "/ipfs/qmbwqxbekc3p8tqskc98xmwnzrzdtrlmimpl8wbutgsmnr/": invalid CID: selected encoding not supported: This looks like a CIDv0 that has been lowercased. Convert the CIDv0 to CIDv1 in case-insensitive base32 using 'ipfs cid base32 <Qm..>' and try again.`

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -254,6 +254,9 @@ func Decode(v string) (Cid, error) {
 
 	_, data, err := mbase.Decode(v)
 	if err != nil {
+		if len(v) == 46 && v[:2] == "qm" { // https://github.com/ipfs/go-ipfs/issues/7792
+			return Undef, fmt.Errorf("%v: This looks like a CIDv0 that has been lowercased. Convert the CIDv0 to CIDv1 in case-insensitive base32 using 'ipfs cid base32 <Qm..>' and try again.", err)
+		}
 		return Undef, err
 	}
 
--- a/cid_test.go
+++ b/cid_test.go
@@ -332,6 +332,17 @@ func TestV0ErrorCases(t *testing.T) {
 	}
 }
 
+func TestV0ErrorDueToLowercase(t *testing.T) {
+	badb58 := "qmbwqxbekc3p8tqskc98xmwnzrzdtrlmimpl8wbutgsmnr"
+	_, err := Decode(badb58)
+	if err == nil {
+		t.Fatal("should have failed to decode")
+	}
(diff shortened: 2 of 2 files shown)


cid.go
```diff
@@ -254,6 +254,9 @@ func Decode(v string) (Cid, error) {
 
 	_, data, err := mbase.Decode(v)
 	if err != nil {
+		if len(v) == 46 && v[:2] == "qm" { // https://github.com/ipfs/go-ipfs/issues/7792
+			return Undef, fmt.Errorf("%v: This looks like a CIDv0 that has been lowercased. Convert the CIDv0 to CIDv1 in case-insensitive base32 using 'ipfs cid base32 <Qm..>' and try again.", err)
```
rvagg, line 258: Seems like an awfully low level to be handing out command specific advice, since this library has wide use, well beyond go-ipfs. This could arise out of use within Filecoin, or a CAR manipulation library, use-cases where `ipfs` isn't even a thing.

Is there a place above the stack from here, closer to the user, where this check could be inserted instead?

In lieu of that, maybe softening it to seem a little less prescriptive?

```
"%v: This may be a CIDv0 that has been lowercased and cannot be decoded. Base32 CIDv1 is recommended where case-insensitivity is required (e.g. with 'ipfs cid base32 <Qm..>')."
```

mvdan: I agree with Rod; the error message here should be reasonably generic, and IPFS could always do the extra check if go-cid returns an error.

jessicaschilling (approved):
I can't weigh in on how generic/specific we need to be, but either the original proposed text or the message in https://github.com/ipfs/go-cid/pull/116#discussion_r533303153 read well to me.

Stebalien:
This is probably best fixed in go-path.

rvagg:
go-path is a much better idea, since it's already ipfs-specific. I've copied this fix, almost exactly as is (including original msg), to https://github.com/ipfs/go-path/pull/33, would that be acceptable to all?

lidel:
I believe https://github.com/ipfs/go-path/pull/33 is fine, this can be closed.
Thank you all  for landing that one!

PR #117, opened 2021-02-28
Extend go-cid to implement the ipld-prime Link interface

willscott:
[Ipld-prime](https://github.com/ipld/go-ipld-prime) uses the term `Link` to describe what a CID does, and uses a [new vocabulary](https://github.com/ipld/go-ipld-prime/blob/8fef5312eb1265f44d42ee4cb2d2bca9dc516327/link.go) for talking about how a Cid and Prefix relate. This adds the requisite methods `Prototype` and `BuildLink` to allow the current implementation to fullfill the IPLD-prime interfaces so that existing CIDs can be directly used as links.


cid.go
```diff
@@ -589,6 +595,13 @@ func (p Prefix) Sum(data []byte) (Cid, error) {
 	}
 }
 
+// BuildLink returns a new Link value based on the given hashsum,
+// and allows Prefix to satisfy the LinkPrototype interface.
+func (p Prefix) BuildLink(hashsum []byte) Cid {
+	c, _ := p.Sum(hashsum)
```
mvdan, line 601: panic if there's an error, instead? otherwise latent bugs could go unnoticed

willscott: I think this does what you want, in that if an invalid - (i guess i'm thinking like `[]byte{}` but even that is valid?) sum is given the response is Cid.Undef which can be matched as an error case still

mvdan: Ah, I see. I think that warrants a mention in the godoc, or at least a comment above the line that ignores the error. If I was reading this function declaration alone, I'd have guessed there was a bug :)

willscott: extended comment

PR #118, opened 2021-03-10
Update go-multihash.  Now uses registry system.

warpfork:
~Some tests are updated to pull in all the hashes that they use, since some of them became optional rather than registered by default. This is what the new side-effecting imports handle.~

~Applications further downstream may require similar imports if they need to use hashes which are no longer registered by default.  (These registrations only happen in *tests* in go-cid, so they're not passed on to downstreams; other applications and libraries must manage their own opt-in.)~

You can't see it from here, but go-mulithash now uses registry system,
so it's reasonably possible to introduce new hashers, and to use
(some parts of!) go-multihash without bringing in lots of transitive
dependencies.

The main package of go-multihash still brings in everything
transitively that it did before, so go-cid's transitives aren't
shrinking, and no code is changing here... but it's closer.
If we did cut over to the new go-mulithash/core, we could make
many transitive dependencies become optional.


warpfork:
This turns out to be fairly inconsequential, because overall, go-cid still uses the `go-multihash` root package, which didn't significantly change in v0.0.15... so, downstreams of go-cid can already pull in the latest versions go-multihash, and things resolve just fine.

But we still might as well merge this update and propagating things proactively.

Stebalien (approved):

Stebalien:
This needs to depend on a released version of go-multihash, but otherwise SGTM.

Stebalien (approved):

PR #122, opened 2021-05-07
chore: fixups from running go vet, go fmt and staticcheck

iand:
There is a race which will be fixed in a following PR.


marten-seemann (approved):

marten-seemann:
@Stebalien Same problem with Travis here as in https://github.com/ipfs/go-namesys/pull/14#issuecomment-835397359.

PR #123, opened 2021-05-10
CI config: update golang version.

warpfork:
This will become relevant when I bump the go-multihash dependency
in a moment, because that repo now uses golang stdlib features
that are from a relatively modern era.

(Go 1.11 is a *long* time ago now.)

This is a fairly boring change, so I'm going to merge it immediately as soon as CI confirms it's in fact fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
 language: go
 
 go:
-  - 1.11.x
+  - 1.15.x
 
 env:
   global:


marten-seemann:
https://github.com/ipfs/go-cid/pull/122 fixed all CI failures, so we should be able to enable the Unified CI Setup today.

warpfork:
> #122 fixed all CI failures, so we should be able to enable the Unified CI Setup today.

Oh, super!  Let's do that instead.

marten-seemann:
Done: #124.
We have the same problem there though: https://github.com/ipfs/go-cid/pull/124/checks?check_run_id=2547504037.

PR #127, opened 2021-07-02
implement CidFromReader

mvdan:
(see commit message)

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -680,3 +680,139 @@ func CidFromBytes(data []byte) (int, Cid, error) {
 
 	return l, Cid{string(data[0:l])}, nil
 }
+
+func toBufByteReader(r io.Reader, dst []byte) *bufByteReader {
+	// If the reader already implements ByteReader, use it directly.
+	// Otherwise, use a fallback that does 1-byte Reads.
+	if br, ok := r.(io.ByteReader); ok {
+		return &bufByteReader{direct: br, dst: dst}
+	}
+	return &bufByteReader{fallback: r, dst: dst}
+}
+
+type bufByteReader struct {
+	direct   io.ByteReader
+	fallback io.Reader
+
+	dst []byte
+}
+
+func (r *bufByteReader) ReadByte() (byte, error) {
+	// The underlying reader has ReadByte; use it.
+	if br := r.direct; br != nil {
+		b, err := br.ReadByte()
+		if err != nil {
+			return 0, err
+		}
+		r.dst = append(r.dst, b)
+		return b, nil
+	}
+
+	// Fall back to a one-byte Read.
+	// TODO: consider reading straight into dst,
+	// once we have benchmarks and if they prove that to be faster.
+	var p [1]byte
+	if _, err := io.ReadFull(r.fallback, p[:]); err != nil {
+		return 0, err
+	}
+	r.dst = append(r.dst, p[0])
+	return p[0], nil
+}
+
+// CidFromReader reads a precise number of bytes for a CID from a given reader.
+// It returns the number of bytes read, the CID, and any error encountered.
+// The number of bytes read is accurate even if a non-nil error is returned.
+//
+// It's recommended to supply a reader that buffers and implements io.ByteReader,
+// as CidFromReader has to do many single-byte reads to decode varints.
+// If the argument only implements io.Reader, single-byte Read calls are used instead.
+func CidFromReader(r io.Reader) (int, Cid, error) {
+	// 64 bytes is enough for any CIDv0,
+	// and it's enough for most CIDv1s in practice.
+	// If the digest is too long, we'll allocate more.
+	br := toBufByteReader(r, make([]byte, 0, 64))
+
+	// We read the first varint, to tell if this is a CIDv0 or a CIDv1.
+	// The varint package wants a io.ByteReader, so we must wrap our io.Reader.
+	vers, err := varint.ReadUvarint(br)
+	if err != nil {
+		return len(br.dst), Undef, err
+	}
+
+	// If we have a CIDv0, read the rest of the bytes and cast the buffer.
+	if vers == mh.SHA2_256 {
+		if n, err := io.ReadFull(r, br.dst[1:34]); err != nil {
+			return len(br.dst) + n, Undef, err
+		}
+
+		br.dst = br.dst[:34]
+		h, err := mh.Cast(br.dst)
+		if err != nil {
+			return len(br.dst), Undef, err
+		}
+
+		return len(br.dst), Cid{string(h)}, nil
+	}
+
+	if vers != 1 {
+		return len(br.dst), Undef, fmt.Errorf("expected 1 as the cid version number, got: %d", vers)
+	}
+
+	// CID block encoding multicodec.
+	_, err = varint.ReadUvarint(br)
+	if err != nil {
+		return len(br.dst), Undef, err
+	}
+
+	// We could replace most of the code below with go-multihash's ReadMultihash.
+	// Note that it would save code, but prevent reusing buffers.
+	// Plus, we already have a ByteReader now.
+	mhStart := len(br.dst)
+
+	// Multihash hash function code.
+	_, err = varint.ReadUvarint(br)
+	if err != nil {
+		return len(br.dst), Undef, err
+	}
+
+	// Multihash digest length.
+	mhl, err := varint.ReadUvarint(br)
+	if err != nil {
+		return len(br.dst), Undef, err
+	}
+
+	// Refuse to make large allocations to prevent OOMs due to bugs.
+	const maxDigestAlloc = 32 << 20 // 32MiB
+	if mhl > maxDigestAlloc {
+		return len(br.dst), Undef, fmt.Errorf("refusing to allocate %d bytes for a digest", mhl)
+	}
+
+	// Fine to convert mhl to int, given maxDigestAlloc.
+	prefixLength := len(br.dst)
+	cidLength := prefixLength + int(mhl)
+	if cidLength > cap(br.dst) {
+		// If the multihash digest doesn't fit in our initial 64 bytes,
+		// efficiently extend the slice via append+make.
+		br.dst = append(br.dst, make([]byte, cidLength-cap(br.dst))...)
+	} else {
+		// The multihash digest fits inside our buffer,
+		// so just extend its capacity.
+		br.dst = br.dst[:cidLength]
+	}
+
+	if n, err := io.ReadFull(r, br.dst[prefixLength:cidLength]); err != nil {
+		// We can't use len(br.dst) here,
+		// as we've only read n bytes past prefixLength.
+		return prefixLength + n, Undef, err
+	}
+
+	// This simply ensures the multihash is valid.
+	// TODO: consider removing this bit, as it's probably redundant;
+	// for now, it helps ensure consistency with CidFromBytes.
+	_, _, err = mh.MHFromBytes(br.dst[mhStart:])
+	if err != nil {
+		return len(br.dst), Undef, err
+	}
+
+	return len(br.dst), Cid{string(br.dst)}, nil
+}
--- a/cid_test.go
+++ b/cid_test.go
@@ -4,10 +4,12 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"math/rand"
 	"reflect"
 	"strings"
 	"testing"
+	"testing/iotest"
 
 	mbase "github.com/multiformats/go-multibase"
 	mh "github.com/multiformats/go-multihash"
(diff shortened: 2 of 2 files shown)


mvdan:
Not super helpful that codecov is complaining about every uncovered line... I'd need to spend another 4h on this to get 100% coverage, and it feels somewhat futile.

cid.go
```diff
@@ -680,3 +680,142 @@ func CidFromBytes(data []byte) (int, Cid, error) {
...
+		return &bufByteReader{direct: br, dst: dst, alreadyRead: len(dst)}
+	}
+	return &bufByteReader{fallback: r, dst: dst, alreadyRead: len(dst)}
+}
+
+type bufByteReader struct {
+	direct   io.ByteReader
+	fallback io.Reader
+
+	alreadyRead int
```
dirkmc, line 697: Suggestion:

I got confused by the naming of `alreadyRead` - I thought it mean "the number of bytes that have already been read from the `bufByteReader`", when actually it means "the number of bytes that were already read from the original reader and passed to `bufByteReader`".

I think for me it would be easier to follow the code if instead we have a variable called `index` or similar that counts up from zero, rather than a variable that counts down from the length of `dst`.

dirkmc (approved):
Logically the code LGTM so I'm approving 👍 

I have one suggestion about how to refactor to make it easier to follow.

cid.go
mvdan: I thought about this for a while too, but I couldn't figure out a nice enough system. Having two separate byte slices might be clearer for the docs, but it made the code more complex, for example.

As for an index - you mean an index that starts at 0, reuses bytes up to len(dst), and then does reads after that?

dirkmc: > you mean an index that starts at 0, reuses bytes up to len(dst), and then does reads after that?

Yes exactly - IIUC that's what the code is doing, but in reverse. In my mind it would be simpler to count up than count down.

mvdan: Sounds good - done. I ended up calling it "consumed", as that felt a bit more descriptive than "index". It also made the constructor a bit simpler.

cid.go
```diff
@@ -680,3 +680,143 @@ func CidFromBytes(data []byte) (int, Cid, error) {
...
+func CidFromReader(r io.Reader) (int, Cid, error) {
+	// 64 bytes is enough for any CIDv0,
+	// and it's enough for most CIDv1s in practice.
+	// If the digest is too long, we'll allocate more.
+	buf := make([]byte, 0, 64)
+
+	// We read two bytes, to tell if this is a CIDv0 or a CIDv1.
+	v0head := buf[:2]
+	if _, err := io.ReadFull(r, v0head[:]); err != nil {
+		return 0, Undef, err
```
raulk, line 744: IIUC, `io.ReadFull` can return an error and yet have consumed bytes. We should propagate that consumed byte count to the caller.

cid.go
```diff
@@ -680,3 +680,143 @@ func CidFromBytes(data []byte) (int, Cid, error) {
...
+	}
+
+	// If we have a CIDv0, read the rest of the bytes and cast the buffer.
+	if v0head[0] == mh.SHA2_256 && v0head[1] == 32 {
+		if _, err := io.ReadFull(r, buf[2:34]); err != nil {
+			return 0, Undef, err
```
raulk, line 750: Propagate `len(buf)` + returned n from `io.ReadFull`? Same for error cases below.

cid.go
mvdan: Good point, returning 0 for all these is a bit lazy. It's what CidFromBytes does, but arguably that doesn't matter, because there it's only useful if err==nil.

I've updated all error returns.

cid.go
mvdan: Done

rvagg (approved):
👌 very nice

Out of interest - is this style of return: `(int, foo, error)` where `int` is always meaningful even if `error` is non-nil an idiomatic Go thing? Would a user naturally assume that the first return value shouldn't be ignored in a non-fatal error? It's good that this is mentioned in the doc at least and I don't suppose there another way around this without a more sophisticated reader interface (like a pushback).

cid.go
```diff
@@ -680,3 +680,145 @@ func CidFromBytes(data []byte) (int, Cid, error) {
...
+	}
+
+	// CID block encoding multicodec.
+	_, err = varint.ReadUvarint(br)
+	if err != nil {
+		return len(br.dst), Undef, err
+	}
+
+	// We could replace most of the code below with go-multihash's ReadMultihash.
+	// Note that it would save code, but prevent reusing buffers.
```
warpfork, line 783: I don't know if I want to encourage you to recurse up into trying to solve this, but I remember stubbing my toe on it previously as well, so I confirm that there's probably improvements that could be made in go-multihash if we chose to go about it that way.

warpfork:
I'd love to have some tests that check that all the byte buffering dancing is correct.  (It _looks_ correct, but I've stopped trusting my eyes and mental model of the compiler completely in things as fiddly as this.)

- benchmarks are always good
- outright counting in a test is also possible -- https://github.com/ipld/go-ipld-prime/blob/b7347f196aa5653fddb5631124f8821ce425bfc9/codec/codectools/scratch/reader_test.go#L49 is an example of doing this

Either would be good.  The latter is more fragile and gives less holistic impact numbers, but also results in something that can actually go red/green in CI rather than requiring someone to keep an eyeball on things, so both have different kinds of value.

I don't know if I consider this a blocking request or not, though.  If it's already been profiled from some place where it's being used, I'll believe it's correct.  Having tests or benchmarks in the same repo will just make it easier to be sure we don't ever regress it.

warpfork:
Interface LGTM, anyway.

Returning an int like this to report partial progress in I/O reading is golang normative, yes.  It describes a side effect that you couldn't otherwise get much insight into, and while that data's not always needed, when it is, you really need it, so it's good practice to always return it.

cid.go
```diff
@@ -680,3 +680,145 @@ func CidFromBytes(data []byte) (int, Cid, error) {
...
+			return len(buf), Undef, err
+		}
+
+		return len(buf), Cid{string(h)}, nil
+	}
+
+	// The varint package wants a io.ByteReader, so we must wrap our io.Reader.
+	// Note that we already read two bytes, so bufByteReader uses those first.
+	// After those two bytes, bufByteReader appends the read bytes to br.dst.
+	br := toBufByteReader(r, buf[:2])
```
Stebalien, line 766: Instead of reading off a few bytes initially and using custom types, let's just start by reading a single varint:

1. If the varint is the CIDv1 multicodec, read as a CIDv1.
2. Otherwise, try to read it as a CIDv0.

cid.go
mvdan: Certainly TODO material for now :) I'm not even sure it would be possible to expose an API that consumes an io.Reader while also reusing a buffer to store the result - that seems extremely clunky and prone to misuse.

mvdan:
> I don't know if I consider this a blocking request or not, though. If it's already been profiled from some place where it's being used, I'll believe it's correct. Having tests or benchmarks in the same repo will just make it easier to be sure we don't ever regress it.

This PR is pure functionality, not optimization/performance. It's virtually impossible to accomplish this without CidFromReader today, unless you're happy with reading a "probably large enough" chunk of bytes and hoping for the best.

In that way, I don't think performance should block this PR. I agree benchmarks and profiling would be nice, but that seems like it should come after. All I can say on the matter is that our use in carv2 works, and seems fast enough :)

I admit I didn't go out of my way to write more tests to try to get 100% test coverage on this func. I'll see if I can reuse `TestBadCidFromBytes` for that purpose.

cid.go
mvdan: Good point; done. Overall saves ~15 lines of code, and the weird double-read dance. I kept the exact same v0-or-v1 logic as CidFromBytes, for the sake of consistency and returning the same kinds of errors. If we think the detection should be different, IMO both should be changed.

cid.go
```diff
@@ -680,3 +680,145 @@ func CidFromBytes(data []byte) (int, Cid, error) {
...
+// The number of bytes read is accurate even if a non-nil error is returned.
+//
+// It's recommended to supply a reader that buffers and implements io.ByteReader,
+// as CidFromReader has to do many single-byte reads to decode varints.
+// If the argument only implements io.Reader, single-byte Read calls are used instead.
+func CidFromReader(r io.Reader) (int, Cid, error) {
+	// 64 bytes is enough for any CIDv0,
+	// and it's enough for most CIDv1s in practice.
+	// If the digest is too long, we'll allocate more.
+	buf := make([]byte, 0, 64)
```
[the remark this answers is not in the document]
mvdan: (now irrelevant thanks to your other comment)

mvdan:
The leading varint simplification that Steven suggested is now done.

> I'll see if I can reuse `TestBadCidFromBytes` for that purpose.

That's now done too.

cid.go
```diff
@@ -680,3 +680,129 @@ func CidFromBytes(data []byte) (int, Cid, error) {
...
+	}
+
+	// Update buf's length.
+	// We're not reading single bytes beyond this point.
+	buf = br.dst
+	br = nil
+
+	// Multihash digest; might be too long, so allocate.
+	// Refuse to make large allocations to prevent OOMs due to bugs.
+	// TODO: reuse buf if it has enough space
```
Stebalien, line 790: Or just always reuse buf but extend the capacity?

```go
prefixLen := len(buf)
if prefixLen+mhl < cap(buf) {
    newBuf := make([]byte, prefixLen+mhl)
    copy(newBuf, buf)
}
io.ReadFull(r, buf[prefixLen:])
```

cid.go
```diff
@@ -680,3 +680,129 @@ func CidFromBytes(data []byte) (int, Cid, error) {
...
+		return len(buf), Undef, fmt.Errorf("refusing to allocate %d bytes for a digest", mhl)
+	}
+	digest := make([]byte, int(mhl))
+	if n, err := io.ReadFull(r, digest); err != nil {
+		return len(buf) + n, Undef, err
+	}
+	buf = append(buf, digest...)
+
+	// This simply ensures the multihash is valid.
+	_, _, err = mh.MHFromBytes(buf[mhStart:])
```
Stebalien, line 802: Does this really buy us anything?

(it shouldn't allocate so that's fine, it just doesn't seem necessary)

Stebalien (approved):
Some nits but otherwise lgtm.

cid.go
mvdan: CidFromBytes uses it, so I went with the same out of consistency.

We read the two varints (code+length) plus the "length" number of digest bytes, so at least we already know the shape is right.

I guess the only extra bit that go-multihash might check is whether the multihash code is valid. If it doesn't, then I agree we can just remove this.

cid.go
```diff
@@ -680,3 +680,129 @@ func CidFromBytes(data []byte) (int, Cid, error) {
...
+		b, err := br.ReadByte()
+		if err != nil {
+			return 0, err
+		}
+		r.dst = append(r.dst, b)
+		return b, nil
+	}
+
+	// Fall back to a one-byte Read.
+	var p [1]byte
```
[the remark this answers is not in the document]
mvdan: I did think about using dst directly, but I didn't want to overcomplicate the code right away without benchmarks. It's a complete guess that it would be faster, I think, given that it's just one byte.

Are you sure that p allocates? We don't have "parse" benchmarks, so it's not easy to tell. From a quick check, I didn't see a "cid.go: p escapes to heap" from `go build -gcflags=-m` on Go 1.16.x.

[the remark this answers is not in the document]
Stebalien: Fair enough to not over complicate now. `r.fallback` is an arbitrary reader so go can't know what `r.fallback.Read` does. So it needs to put `p` on the heap because `r.fallback.Read` could do anything with it.

It's probably not showing up now due to inlining? That's my guess. Given a small enough example, go may be able to determine that `fallback` can _only_ be one specific type of reader.

cid.go
Stebalien: We used to do check codecs, but don't anymore.

But really, this is probably fine to keep.

mvdan: I'll keep it here for now then :) Once we have benchmarks, and if someone cares enough about the penalty, they can make an argument to remove it and do it.

cid.go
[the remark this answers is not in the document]
mvdan: I'm going to add this to "can be optimized once we have benchmarks" :) For now, this is easy, and escape analysis doesn't say it escapes when compiling the func, so... I think we're agreeing to leave as is.

cid.go
```diff
@@ -680,3 +680,129 @@ func CidFromBytes(data []byte) (int, Cid, error) {
...
+	br := toBufByteReader(r, buf)
+	vers, err := varint.ReadUvarint(br)
+	if err != nil {
+		return len(br.dst), Undef, err
+	}
+
+	// If we have a CIDv0, read the rest of the bytes and cast the buffer.
+	if vers == mh.SHA2_256 {
+		if n, err := io.ReadFull(r, buf[1:34]); err != nil {
+			return len(buf) + n, Undef, err
```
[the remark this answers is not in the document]
mvdan: Done - also shaves off six lines of code, since we had to sync the two via `buf = br.dst`.

cid.go
mvdan: Yep, that's what the TODO was for - I was just cutting corners again for the sake of getting this to work. Done.

mvdan:
@Stebalien the two byte slice comments are addressed, and the two perf nits now have TODOs. Want to take one last look?

Stebalien (approved):

PR #128, opened 2021-07-16
amend the CidFromReader slice extension math

mvdan:
(see commit message)

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -793,7 +793,7 @@ func CidFromReader(r io.Reader) (int, Cid, error) {
 	if cidLength > cap(br.dst) {
 		// If the multihash digest doesn't fit in our initial 64 bytes,
 		// efficiently extend the slice via append+make.
-		br.dst = append(br.dst, make([]byte, cidLength-cap(br.dst))...)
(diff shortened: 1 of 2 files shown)


aarshkshah1992 (approved):

cid.go
```diff
@@ -793,7 +793,7 @@ func CidFromReader(r io.Reader) (int, Cid, error) {
 	if cidLength > cap(br.dst) {
 		// If the multihash digest doesn't fit in our initial 64 bytes,
 		// efficiently extend the slice via append+make.
-		br.dst = append(br.dst, make([]byte, cidLength-cap(br.dst))...)
+		br.dst = append(br.dst, make([]byte, cidLength-len(br.dst))...)
```
masih, line 796: would it make sense to reuse `prefixLength` here?

mvdan: Hmmm. I think I prefer leaving it as `len` because it's clearer that we're appending to get `len` to get to `cidLength`. In these three lines of code, the only variable should be the final length/capacity we want, i.e. `cidLength`.

mvdan:
Merging since this is a tiny fix with a test, and a panic is bad and blocking Ignite.

PR #132, opened 2021-09-07
avoid double alloc in NewCidV1

mvdan:
(see commit message)

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -22,6 +22,7 @@ package cid
 import (
 	"bytes"
 	"encoding"
+	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


ribasushi (approved):
This looks sane.

Although I wonder: `Multihash` is a `[]byte`, and `PutUvarint` woks on `[]byte` as well. Would it not be even faster to:
- stack-allocate something reasonable ( ~80-something bytes, you need to account for obscenely large codecs+512bit hashes )
- if stuff fits - work on the above, otherwise `make()` the right size
- write the 1 + uvarint + `copy` the mh
- `string()` the entire resulting thing

❓

mvdan:
Possibly? I guess it's a tradeoff between "copy uvarint bytes around" and "use more stack space and possibly allocate if our stack space isn't big enough". It's hard to make the call without stats. I think that copying the PutUvarint bytes (most likely just a few bytes, max 10) is cheap enough that we shouldn't pay any attention to it :) The new NewCidV1 is faster than the old one, anyway.

Stebalien (approved):

PR #134, opened 2021-09-24
Implement efficient function to check for `IDENTITY` multihash code

### LINKED ISSUE #133: Provide an efficient API to check whether a CID has `IDENTITY` multihash code
CIDs with multihash code `IDENTITY` typically require special handling when encountered in blockstores. This is because, such CIDs contain the data within themselves; the data is simply the multihash digest of that CID, since multihash code `IDENTITY` corresponds to `copy` hash function.

To handle them gracefully checks are needed to indicate whether a given CID has `IDENTITY` code or not, and checks would have to run for almost all operations on blockstore API. It is therefore, highly desirable to check as efficiently as possible. 

The current APIs offered provide two ways to perform the check:
1. [`cid.Prefix().MhType`](https://pkg.go.dev/github.com/ipfs/go-cid#Cid.Prefix)
2. decode of `cid.Hash()` via `go-multihash` API to extract the code

Blockstore implementations would benefit from an API that checks whether a given CID or digest of a CID has `IDENTITY` code in a "fail-fast" manner. This is where the check would return as fast as possible if a CID is not an `IDENTITY` without checking for the validity of the CID first, _then_ decoding digest, _then_ comparing multihash code.

The rationale for a "fail-fast" check is:
1. if a CID does not have `IDENTITY` multihash code, it doesn't always need to be fully decoded in order for a block to be returned (e.g. when CID is used as key in a map)
2. the majority of CIDs interacted with are not `IDENTITY` therefore we want to pay the price of decoding only when we have to, and certainly not for every call to blockstore.

I therefore propose to:
- [x] Write benchmarks that compare the efficiency of the current APIs when checking for `IDENTITY` code.
- [x] Provide an alternative API that aims to improve efficiency for the checks.

masih:
Implement a fail-fast function that checks whether the code of a CID
is `multihash.IDENTITY` or not.

Add benchmarks that compare three ways of checking for
`multihash.IDENTITY` code:
1. `Cid.Prefix().MhType`
2. Decode of `Cid.Has()`
3. The new `Cid.IsIdentity()` API

```
BenchmarkCidV1IdentityCheckUsingPrefix
BenchmarkCidV1IdentityCheckUsingPrefix-8            	355396287	         3.334 ns/op	332899.32 MB/s	       0 B/op	       0 allocs/op
BenchmarkCidV1IdentityCheckUsingMultihashDecode
BenchmarkCidV1IdentityCheckUsingMultihashDecode-8   	 6404062	       194.2 ns/op	5715.67 MB/s	    1216 B/op	       2 allocs/op
BenchmarkCidV1IdentityCheckUsingIsIdentity
BenchmarkCidV1IdentityCheckUsingIsIdentity-8        	442939464	         2.621 ns/op	423535.07 MB/s	       0 B/op	       0 allocs/op
```

Fixes #133

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -26,6 +26,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	multihash "github.com/multiformats/go-multihash/core"
 	"io"
 	"strings"
 
@@ -824,4 +825,36 @@ func CidFromReader(r io.Reader) (int, Cid, error) {
 	}
 
 	return len(br.dst), Cid{string(br.dst)}, nil
+}
+
+// IsIdentity checks whether a CID has multihash.IDENTITY code.
+// Note that this function is optimized to return fast.
+// Checking for multihash.IDENTITY code in this function does not fully verify the validity of the CID.
+func (c Cid) IsIdentity() (bool, error) {
+
+	if c.Version() == 0 {
+		return false, nil
+	}
+
+	var n int
+	// Skip version.
+	_, n, err := uvarint(c.str[n:])
+	if err != nil {
+		return false, err
+	}
+	n += n
+
+	// Skip codec.
+	_, n, err = uvarint(c.str[n:])
+	if err != nil {
+		return false, err
+	}
+	n += n
+
+	// Read multihash code.
+	mhCode, _, err := uvarint(c.str[n:])
+	if err != nil {
+		return false, nil
+	}
+	return mhCode == multihash.IDENTITY, nil
 }
--- a/benchmark_test.go
+++ b/benchmark_test.go
@@ -1,0 +1,83 @@
+package cid_test
+
+import (
+	"github.com/ipfs/go-cid"
+	"github.com/multiformats/go-multihash"
+	"math/rand"
+	"testing"
+)
+
+// randSourceSeed represents the static RNG seed, fixed for reproducibility.
+const randSourceSeed = 1413
+
+// BenchmarkCidV1IdentityCheckUsingPrefix benchmarks checking whether a CIDv1 has multihash.IDENTITY
+// code via. Cid.Prefix().
+func BenchmarkCidV1IdentityCheckUsingPrefix(b *testing.B) {
+	rng := rand.New(rand.NewSource(randSourceSeed))
+	cv1 := generateCidV1(b, rng, multihash.IDENTITY)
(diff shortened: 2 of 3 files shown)


cid.go
```diff
@@ -825,3 +826,35 @@ func CidFromReader(r io.Reader) (int, Cid, error) {
...
+// Checking for multihash.IDENTITY code in this function does not fully verify the validity of the CID.
+func (c Cid) IsIdentity() (bool, error) {
+
+	if c.Version() == 0 {
+		return false, nil
+	}
+
+	var n int
+	// Skip version.
+	_, n, err := uvarint(c.str[n:])
```
marten-seemann, line 841:
```suggestion
	_, n, err := uvarint(c.str)
```

cid.go
```diff
@@ -825,3 +826,35 @@ func CidFromReader(r io.Reader) (int, Cid, error) {
...
+// IsIdentity checks whether a CID has multihash.IDENTITY code.
+// Note that this function is optimized to return fast.
+// Checking for multihash.IDENTITY code in this function does not fully verify the validity of the CID.
+func (c Cid) IsIdentity() (bool, error) {
+
+	if c.Version() == 0 {
+		return false, nil
+	}
+
+	var n int
```
marten-seemann, line 839: No need to declare `n` here.

cid.go
```diff
@@ -825,3 +826,35 @@ func CidFromReader(r io.Reader) (int, Cid, error) {
...
+	// Skip version.
+	_, n, err := uvarint(c.str[n:])
+	if err != nil {
+		return false, err
+	}
+	n += n
```
marten-seemann, line 845: Did you mean do declare a `bytesRead int`, and add `n` every time you read a varint? Now you're just overwriting `n` every time.

Stebalien:
We _really_ shouldn't be special-casing "identity" CIDs at this layer. If we really need this for performance, we should have a function to get the multihash codec.

But I'm not convinced it matters. We're saving less than a nanosecond per op, which'll be nothing compared to all the other work we're going to end up doing.

masih:
Thank you @marten-seemann and @Stebalien for the reviews  🍻 
I agree with @Stebalien.

Closing this as it does not add much value.

masih:
@marten-seemann @Stebalien I have extracted benchmarks from this PR into a new PR as a way to document the efficiency of existing APIs for `IDENTITY` check. I think that's worth having 🙂 

See: #135

PR #135, opened 2021-09-27
Benchmark existing ways to check for `IDENTITY` CIDs

masih:
Benchmark existing ways to check for `IDENTITY` CIDs

Add benchmarks that compare two ways of checking for
`multihash.IDENTITY` code:
1. `Cid.Prefix().MhType`
2. Decode of `Cid.Hash()`

This benchmark illustrates that using Cid.Prefix is more efficient than
`multihash.Decode`. Users wishing to perform such a check should use
`Cid.Prefix`.

Consider that `Cid.Prefix` is already efficient enough and introducing a
dedicated API for performing this check will likely result in small
gains.

```
BenchmarkIdentityCheck
BenchmarkIdentityCheck/Prefix
BenchmarkIdentityCheck/Prefix-8         	351656472	         3.280 ns/op
BenchmarkIdentityCheck/MultihashDecode
BenchmarkIdentityCheck/MultihashDecode-8         	 5715622	       239.1 ns/op
```

Relates to #133

### DIFF
--- a/benchmark_test.go
+++ b/benchmark_test.go
@@ -1,0 +1,61 @@
+package cid_test
+
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/ipfs/go-cid"
+	"github.com/multiformats/go-multihash"
+)
+
+// BenchmarkIdentityCheck benchmarks two ways of checking whether a CIDv1 has multihash.IDENTITY
+// code:
+//  1. Cid.Prefix(), and
+//  2. decoding the Cid.Hash().
+//
+// This benchmark illustrates that using Cid.Prefix is more efficient than multihash.Decode.
+// Users wishing to perform such a check should use Cid.Prefix.
+//
+// Consider that `Cid.Prefix` is already efficient enough and introducing a dedicated API for
+// performing this check will likely result in small gains.
+func BenchmarkIdentityCheck(b *testing.B) {
+	rng := rand.New(rand.NewSource(1413))
+
+	data := make([]byte, rng.Intn(100)+1024)
+	if _, err := rng.Read(data); err != nil {
+		b.Fatal(err)
+	}
+	mh, err := multihash.Sum(data, multihash.IDENTITY, -1)
+	if err != nil {
+		b.Fatal(err)
+	}
+	cv1 := cid.NewCidV1(cid.Raw, mh)
+
+	b.SetBytes(int64(cv1.ByteLen()))
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	b.Run("Prefix", func(b *testing.B) {
+		b.RunParallel(func(pb *testing.PB) {
+			for pb.Next() {
+				if cv1.Prefix().MhType != multihash.IDENTITY {
+					b.Fatal("expected IDENTITY CID")
+				}
+			}
+		})
+	})
+
+	b.Run("MultihashDecode", func(b *testing.B) {
+		b.RunParallel(func(pb *testing.PB) {
+			for pb.Next() {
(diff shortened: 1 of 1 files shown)


mvdan (commented):
You can drop the `CidV1` part of the benchmark name; Cid is redundant since this module is go-cid, and v1 is redundant as we don't really care if the CID is v1 or not - it just so happens that all identity CIDs are currently v1. If they also supported a future CIDv2, then the benchmark would simply do both in each iteration.

You could also make the two methods sub-benchmarks, to simplify sharing code and make the output nicer. For example:

```
func BenchmarkIdentityCheck(b *testing.B) {
    // cid setup, ReportAllocs, etc
    b.Run("Prefix", func(b *testing.B) {
        b.RunParallel(func(pb *testing.PB) {
            // check via Prefix
        })
    })
    b.Run("MultihashDecode", func(b *testing.B) {
        b.RunParallel(func(pb *testing.PB) {
            // check via MultihashDecode
        })
    })
}
```

benchmark_test.go
```diff
@@ -0,0 +1,62 @@
...
+func BenchmarkIdentityCheck(b *testing.B) {
+	rng := rand.New(rand.NewSource(1413))
+	cv1 := generateCidV1(b, rng, multihash.IDENTITY)
+
+	b.SetBytes(int64(cv1.ByteLen()))
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	b.Run("Prefix", func(bb *testing.B) {
+		bb.RunParallel(func(pb *testing.PB) {
```
mvdan, line 21: I wouldn't name the inner one `bb`, because then it's too easy to make the mistake of calling methods on the wrong (parent) `b`, just like you did on line 24. If this sub-benchmark fails, only it should fail, not the entire parent benchmark.

benchmark_test.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		b.Fatal(err)
+	}
+	return mh
+}
+
+func generateCidV1(b *testing.B, rng *rand.Rand, mhCode uint64) cid.Cid {
+	mh := generateMultihash(b, rng, mhCode)
+	return cid.NewCidV1(cid.Raw, mh)
+}
```
mvdan, line 62: consider inlining generateCidV1, or even both funcs, now that they are only called in one place.

benchmark_test.go
```diff
@@ -0,0 +1,62 @@
+package cid_test
+
+import (
+	"github.com/ipfs/go-cid"
+	"github.com/multiformats/go-multihash"
+	"math/rand"
+	"testing"
```
mvdan, line 7: keep the std imports separate :)

mvdan (approved):
SGTM! Probably wait for a review from a steward though.

benchmark_test.go
mvdan: This is one example where shadowing is a good thing: it prevents using the wrong "level".

warpfork (approved):
I have no reason to complain about more benchmarks :)

If you still have the context in your head: I wouldn't mind a short paragraph of comment attached to the function about what you were looking for / why it mattered / and what the rough findings were at the time you committed.  (It'll get stale, sure -- but that doesn't mean it won't still be useful context clues to some future archaeologist^W reviewer.)

masih:
Will do 🍻 

Context captured [here](https://github.com/ipfs/go-cid/issues/133)

TLDR; there are multiple ways of inferring if a CID is `IDENTITY` and one of them is more efficient. The one that's more efficient is efficient enough to avoid the need to implement new APIs to do this (see [closed PR](https://github.com/ipfs/go-cid/pull/134) for more details)

PR #137, opened 2022-03-23
refactor: remove invalid string2code mappings

lidel:
## Problem: string2codec maps are a mess

The current state is not sustainable, string mappings are invalid or missing, and we don't have correct ones for core IPLD formats like `dag-pb`, `dag-json` and `dag-cbor` :see_no_evil: 

Solution: cleanup

## Problem 2: cleanup may introduce silent bugs into people's codebases

https://github.com/ipfs/go-cid/pull/137/commits/da5524780a4ae73d6b2ca765f1df8e7e8b7b0bcd  demonstrates the map changes we would have to do to restore sanity:  fixes mapping, and adds IPLD formats along with non-dag generic variants for Protobuf, JSON and   CBOR, so go-cid finally match the [source of truth](https://github.com/multiformats/multicodec/blob/master/table.csv).

This approach would change the meaning of code `0x70` and `0x71` and strings `cbor` and `protobuf`:

- `protobuf` string now correctly points at code `0x50`  (was `0x70`, which is  `dag-pb`)
  -  `0x70` code is mapped to `dag-pb` (was `protobuf` before)
- `cbor` string now correctly points at code `0x51` (was `0x71`, which is  `dag-cbor`)
  - `0x71` code is now mapped to `dag-cbor` (was `cbor` before)
  
Pretty risky – people dont read release notes, this could cause silent errors in people's software.
Also, we have proper code2string mappings in  [go-multicodec](https://github.com/multiformats/go-multicodec)  (generated, and not written by hand)

# Solution

Remove invalid/redundant string2code mappings from go-ipfs, point ecosystem at using go-multicodec directly.

# BREAKING CHANGE
  
  
## Implemented Approach B: Remove invalid mappings entirely :green_heart: 

I applied this in https://github.com/ipfs/go-cid/pull/137/commits/6ca2617460d553e1aa10d4fdced1dcfedd300197

We already have [go-multicodec](https://github.com/multiformats/go-multicodec) which is generated from the [source of truth](https://github.com/multiformats/multicodec/blob/master/table.csv).

We remove hardcoded maps, then people will be forced to refactor their code to use  go-multicodec (https://github.com/multiformats/go-multicodec/issues/65) – forcing people to refactor makes it less likely they will just blindly bump the dependency without reading release notes.

----

# Refactor guide

Converting  code to string and vice versa will no longer be done with go-cid, go-multicodec should be used instead

## Old  way (go-cid)

```go

cid "github.com/ipfs/go-cid"

// string → code
code := cid.Codecs["libp2p-key"]

// code → string
string := cid.CodecToStr[code]
```

## New way (go-multicodec)

```go
mc "github.com/multiformats/go-multicodec"

// string → code
var mc.Code code 
code.Set("libp2p-key")

// code → string
code.String()
```

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -58,15 +58,9 @@ const (
 	DagJSON     = 0x0129 // https://ipld.io/docs/codecs/known/dag-json/
 	Libp2pKey   = 0x72   // https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#peer-ids
 
-	// generic, non-dag variants
-	Protobuf = 0x50
-	CBOR     = 0x51
-	JSON     = 0x0200
-
 	// other
 	GitRaw                = 0x78
 	DagJOSE               = 0x85 // https://ipld.io/specs/codecs/dag-jose/spec/
-	DagCOSE               = 0x86
 	EthBlock              = 0x90
 	EthBlockList          = 0x91
 	EthTxTrie             = 0x92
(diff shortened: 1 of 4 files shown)


lidel:
I went with Approach B (Remove mappings).
Filled https://github.com/multiformats/go-multicodec/issues/65 to figure out the best way existing users could migrate.

lidel:
Based on https://github.com/multiformats/go-multicodec/issues/65#issuecomment-1078919431 I wrote migration guide (will include it in release notes). 

This is now ready to merge as soon I get some reviews.

mvdan (approved):
SGTM with the understanding that we're okay with the breaking change. An alternative would be to have go-cid import the new go-multicodec, filling the global map at init time - it would avoid breaking users, but longer term it's a worse approach.

If you're breaking users, have you thought about removing the constants and pointing users towards go-multicodec directly? It might be a bit too much in terms of breakage, but if we're already replacing the from/to string APIs, replacing the constants as well would seem consistent.

aschmahmann:
@mvdan the reason to break users with the maps but not the constants is because the maps are **wrong**. The most egregious issue is that `"cbor"` is mapped to `dag-cbor` (0x71) instead of to `cbor` (0x51). There's also the `protobuf` vs `dag-pb` issue, but that's less terrible. If we just fixed the maps then people would silently break which is IMO the worst of the options.

> If you're breaking users, have you thought about removing the constants and pointing users towards go-multicodec directly?

Why would you choose to break users when you could just not at trivial cost? We could add Deprecated markers pointing people at go-multicodec though.

> It might be a bit too much in terms of breakage, but if we're already replacing the from/to string APIs, replacing the constants as well would seem consistent.

With some basic searching:
- https://grep.app/search?current=11&q=cid.Raw&words=true has 110 results
- https://grep.app/search?q=cid.Codecs&words=true has 6
- https://grep.app/search?q=cid.CodecToStr&words=true has 11

So breaking the maps seems to cause much less trouble for people than also breaking the constants.

cid.go
```diff
@@ -51,15 +51,16 @@ var (
...
-	Raw = 0x55
-
-	DagProtobuf = 0x70
-	DagCBOR     = 0x71
-	Libp2pKey   = 0x72
-
-	GitRaw = 0x78
-
-	DagJOSE               = 0x85
+	// core IPLD
```
aschmahmann, line 54: I don't think I'd differentiate in comments "core IPLD" from other codecs.

That being said I think the groupings are fine as they highlight the ones developers are currently most likely to see in the wild.

mvdan:
ACK, I hadn't caught the detail about the bad map entries.

cid.go
```diff
@@ -51,15 +51,16 @@ var (
...
+	Raw         = 0x55
```
aschmahmann, line 55: Maybe it's worth deprecating these and pointing to https://github.com/multiformats/go-multicodec even though we're going to leave them in.

aschmahmann (approved):
LGTM. While generally, it'd be nice to see some PRs in the downstream breakages (e.g. go-ipfs) to see how this plays out I suspect it's fine here to merge. This repo doesn't get too many changes so it's fair to assume we won't make more before we deal with the breakages in go-ipfs and related repos and we can figure this out later.

Whether you want to tag now, before we have PRs resolving the breakages in the few effected projects is your call.

cid.go
rvagg: yeah, although `Libp2pKey` isn't "core IPLD", probably not a big deal though

rvagg (approved):

lidel:
note to self: before I merge this, I'll test this in go-ipfs (I'll do that as part of review of https://github.com/ipfs/go-ipfs/pull/8568)

lidel:
Added deprecation notice and updated README in https://github.com/ipfs/go-cid/pull/137/commits/85198b907298a7071d2b5fa0405d92e5eb8aecec

lidel:
This is unblocked and ok to merge:

- Both resource manager (https://github.com/ipfs/go-ipfs/pull/8680) and multicodec cleanup in `block put` and `cid codecs` commands (https://github.com/ipfs/go-ipfs/pull/856)8 landed in go-ipfs master. 
- Improved README landed in https://github.com/multiformats/go-multicodec/pull/66 


**TLDR**: if you've found this PR because you need to map CODE to STRING or STRING TO CODE, see usage https://github.com/multiformats/go-multicodec#usage

PR #139, opened 2022-06-24
Add MustParse

mg98:
I found this function to be extremely useful, especially for testing.


Stebalien (approved):
I've seen/used this function several times as well.

PR #140, opened 2022-07-01
adding blake3 support, bumping versions

laudiacay:
adding blake3 stuff and bumping multihash version. bumping minor version here too.


laudiacay:
depends on this getting merged: https://github.com/multiformats/go-multihash/pull/158

Jorropo:
Golang's dependency model is not tree like.

Except counting `/v2`, `/v3`, ... branches (which we are not using here). Only one version of the package is compiled (the latest one).

If a consumer (like `go-ipfs`) updates to a later `go-multihash` this also updates it for all other users of this module (like `go-cid`).

I think this can be closed, if someone need some custom `go-multihash` for some blake3 thing they can just update to the newer  `go-multihash`.

laudiacay:
phew one less to worry about.... so i guess we just need to bump go-verifcid on go-ipfs, and get the go-multihash bugfix in.

PR #141, opened 2022-08-18
fix: return nil Bytes() if the Cid in undef.

Jorropo:


Stebalien (approved):

Stebalien:
LGTM, but could you document this before merging?

PR #142, opened 2022-08-22
fix: bring back, but deprecate CodecToStr and Codecs

rvagg:
Uses go-multicodec as the source of truth this time.

So ... I think I want to assert that we made a mistake in https://github.com/ipfs/go-cid/pull/137 by removing it entirely. It continues to be a source of pain across our whole Go ecosystem of packages - primarily coming from the github.com/libp2p/go-libp2p-core usage of it in its `FromCid()`. Now, upgrading dependencies means following up a whole chain of (often unrelated) dependencies just to get libp2p upgraded. My latest instance of this is in trying to get some basic deps upgraded in go-merkledag, which shouldn't care about libp2p but does for some transitive dependencies in some test things. Ultimately that goes back to needing to upgrade go-bitswap's libp2p dependencies .. which I really don't want to have to go just to get go-merkledag upgraded.

This time, we're deferring to go-multicodec for the mappings, and I'd like to get dependabot setup here too to make sure we keep that updated. There's also a deprecation notice which at least staticcheck should pick up.

I think, in lieu of proper semver-major semantics, this is playing like a good open source semver citizen, moreso than just yanking it out? Maybe next time for something like this we can bite the bullet and semver-major bump.

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -30,6 +30,7 @@ import (
 	"strings"
 
 	mbase "github.com/multiformats/go-multibase"
+	"github.com/multiformats/go-multicodec"
 	mh "github.com/multiformats/go-multihash"
 	varint "github.com/multiformats/go-varint"
 )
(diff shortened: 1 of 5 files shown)


aschmahmann:
@rvagg IIRC the problem with just changing the values in the map (as done here) is that some of the map strings have changed in the process and there's no compiler-based indicators of the changes to the names used for dag-pb and dag-cbor, particularly dag-cbor where the previous name here "cbor" now means a different codec.

While the removal has some pain associated with it, at least it's explicit and users will notice it as compared to just changing the functionality and deprecating it.

If you need some help upgrading go-libp2p in go-bitswap I'd bet @Jorropo or @guseggert could help you out.

rvagg:
> IIRC the problem with just changing the values in the map (as done here) is that some of the map strings have changed

Yep, that was the root problem, but maybe we should have deprecated rather than ripped them out. But also, the main problem is not `Codecs` but `CodecToStr` which goes the other way, where you usually want a user-printable form, which is reasonable, and is often not critical if it changes and all that code. In the libp2p instance it's simply to print a nicer error message: https://github.com/libp2p/go-libp2p-core/blob/243f8b9e3c8185705c55510a8c15c65abee739f6/peer/peer.go#L167-L171 - and now, we've replaced the exact behaviour but with go-multicodec: https://github.com/libp2p/go-libp2p/blob/master/core/peer/peer.go#L160-L162 printing a string form of a codec! So what we've gained is pushing people to a library with a more comprehensive, and generated list of codecs, great, but we've needlessly broken the ecosystem in the process when we could have just made a redirection from here to there with the same behaviour.

I could remove the `Codecs` map, the value of that is a bit more dubious. But I still think `Deprecated` is a better way of dealing with this when we're not doing proper (>0) semver.

Jorropo (approved):
I need this to update Kubo to go-libp2p v0.22.0 (I could also fix the dep but I'll see this later) LGTM.

PR #145, opened 2022-09-04
Revert #142

### LINKED ISSUE #144: CodecToStr and Codecs in v0.3.1 could cause data corruption
[v0.2.0](https://github.com/ipfs/go-cid/releases/tag/v0.2.0) we made a conscious decision to make a breaking change to FORCE people to refactor their code, and not be surprised by code changes.  Rationale in https://github.com/ipfs/go-cid/pull/137.

In golang major  bump is essentially a different package – bumping major does not help people who are already using INVALID mappings.  That is why we did it in v0.2. 

Recently merged PR https://github.com/ipfs/go-cid/pull/142 removed that safety and could cause unexpected code change if someone updates from v1.0 to v3.0. IIUC we now have this:

- What if someone used `cbor` string and expected it to point at `0x71` (dag-cbor)? 
  - v0.3.1  SILENTLY changes the mapping to `0x51` (cbor)
- What if someone used `protobuf`   string and expected it to point at`0x70` (dag-pb)
  - v0.3.1 SILENTLY changes the mapping to `0x50` (protobuf)

Upgrading v0.1.x to  v0.3.1 can now produce DAGs with different codec and get NO warning about this BREAKING CHANGE.

@Jorropo @rvagg If you reallly want to keep CodecToStr and Codecs, you should make sure they return hard error when someone tries to use them for impacted mappings above.  Third-party people use this library. People don't read release notes. Silent data corruption is not acceptable.

Jorropo:
Revert #142
Fixes #144 

> @Jorropo @rvagg If you reallly want to keep CodecToStr and Codecs, you should make sure they return hard error when someone tries to use them for impacted mappings above. Third-party people use this library. People don't read release notes. Silent data corruption is not acceptable.

I don't think this is doable other than removing CodecToStr.

An other solution is patching go-cid's map to present the old behaviour.

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -30,7 +30,6 @@ import (
 	"strings"
 
 	mbase "github.com/multiformats/go-multibase"
-	"github.com/multiformats/go-multicodec"
 	mh "github.com/multiformats/go-multihash"
 	varint "github.com/multiformats/go-varint"
 )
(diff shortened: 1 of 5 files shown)


Jorropo:
I know this is a semver violation but no one used v0.3.1 already so it's fine.

lidel (approved):
revert as a quick fix to unblock Kubo sgtm.

Jorropo:
For kubo Ill have to fix something to use go multicodec but that fine.

PR #146, opened 2022-10-18
edit README.md example error

susarlanikhilesh:
Codec in cid.Prefix expects uint64 but the assignment has multicodec.Code type


rvagg (approved):
thanks!

PR #150, opened 2023-03-15
feat: wrap parsing errors into ErrInvalidCid

hacdias:
This PR wraps the parsing errors into `ErrInvalidCid`, a new custom error type I added. This allows depends of this package to test if the error they got includes a wrapped CID parsing error via `errors.Is`.

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -37,10 +37,32 @@ import (
 // UnsupportedVersionString just holds an error message
 const UnsupportedVersionString = "<unsupported cid version>"
 
+// ErrInvalidCid is an error that indicates that a CID is invalid.
+type ErrInvalidCid struct {
+	Err error
+}
+
+func (e ErrInvalidCid) Error() string {
+	return fmt.Sprintf("invalid cid: %s", e.Err)
+}
+
+func (e ErrInvalidCid) Unwrap() error {
+	return e.Err
+}
+
+func (e ErrInvalidCid) Is(err error) bool {
+	switch err.(type) {
+	case ErrInvalidCid, *ErrInvalidCid:
+		return true
+	default:
+		return false
+	}
+}
+
 var (
 	// ErrCidTooShort means that the cid passed to decode was not long
 	// enough to be a valid Cid
-	ErrCidTooShort = errors.New("cid too short")
+	ErrCidTooShort = ErrInvalidCid{errors.New("cid too short")}
 
 	// ErrInvalidEncoding means that selected encoding is not supported
 	// by this Cid version
@@ -90,10 +112,10 @@ func tryNewCidV0(mhash mh.Multihash) (Cid, error) {
 	// incorrectly detect it as CidV1 in the Version() method
 	dec, err := mh.Decode(mhash)
 	if err != nil {
-		return Undef, err
+		return Undef, ErrInvalidCid{err}
 	}
 	if dec.Code != mh.SHA2_256 || dec.Length != 32 {
-		return Undef, fmt.Errorf("invalid hash for cidv0 %d-%d", dec.Code, dec.Length)
+		return Undef, ErrInvalidCid{fmt.Errorf("invalid hash for cidv0 %d-%d", dec.Code, dec.Length)}
 	}
 	return Cid{string(mhash)}, nil
 }
@@ -177,7 +199,7 @@ func Parse(v interface{}) (Cid, error) {
 	case Cid:
 		return v2, nil
 	default:
-		return Undef, fmt.Errorf("can't parse %+v as Cid", v2)
+		return Undef, ErrInvalidCid{fmt.Errorf("can't parse %+v as Cid", v2)}
 	}
 }
 
@@ -210,7 +232,7 @@ func Decode(v string) (Cid, error) {
 	if len(v) == 46 && v[:2] == "Qm" {
 		hash, err := mh.FromB58String(v)
 		if err != nil {
-			return Undef, err
+			return Undef, ErrInvalidCid{err}
 		}
 
 		return tryNewCidV0(hash)
@@ -218,7 +240,7 @@ func Decode(v string) (Cid, error) {
 
 	_, data, err := mbase.Decode(v)
 	if err != nil {
-		return Undef, err
+		return Undef, ErrInvalidCid{err}
 	}
 
 	return Cast(data)
@@ -240,7 +262,7 @@ func ExtractEncoding(v string) (mbase.Encoding, error) {
 	// check encoding is valid
 	_, err := mbase.NewEncoder(encoding)
 	if err != nil {
-		return -1, err
+		return -1, ErrInvalidCid{err}
 	}
 
 	return encoding, nil
@@ -260,11 +282,11 @@ func ExtractEncoding(v string) (mbase.Encoding, error) {
 func Cast(data []byte) (Cid, error) {
 	nr, c, err := CidFromBytes(data)
 	if err != nil {
-		return Undef, err
+		return Undef, ErrInvalidCid{err}
 	}
 
 	if nr != len(data) {
-		return Undef, fmt.Errorf("trailing bytes in data buffer passed to cid Cast")
+		return Undef, ErrInvalidCid{fmt.Errorf("trailing bytes in data buffer passed to cid Cast")}
 	}
 
 	return c, nil
@@ -434,7 +456,7 @@ func (c Cid) Equals(o Cid) bool {
 // UnmarshalJSON parses the JSON representation of a Cid.
 func (c *Cid) UnmarshalJSON(b []byte) error {
 	if len(b) < 2 {
-		return fmt.Errorf("invalid cid json blob")
+		return ErrInvalidCid{fmt.Errorf("invalid cid json blob")}
 	}
 	obj := struct {
 		CidTarget string `json:"/"`
@@ -442,7 +464,7 @@ func (c *Cid) UnmarshalJSON(b []byte) error {
 	objptr := &obj
 	err := json.Unmarshal(b, &objptr)
 	if err != nil {
-		return err
+		return ErrInvalidCid{err}
 	}
 	if objptr == nil {
 		*c = Cid{}
@@ -450,12 +472,12 @@ func (c *Cid) UnmarshalJSON(b []byte) error {
 	}
 
 	if obj.CidTarget == "" {
-		return fmt.Errorf("cid was incorrectly formatted")
+		return ErrInvalidCid{fmt.Errorf("cid was incorrectly formatted")}
 	}
 
 	out, err := Decode(obj.CidTarget)
 	if err != nil {
-		return err
+		return ErrInvalidCid{err}
 	}
 
 	*c = out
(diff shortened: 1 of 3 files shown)


rvagg (commented):
a test or two would be nice

Is the custom `Is` necessary here? The `Unwrap` should be enough for `errors.Is` to work shouldn't it?

It's a shame we're not at minimum of Go 1.20 because `fmt.Errorf("%w: %w", ErrInvalidCid, err)` would be an excellent pattern to avoid the additional code here, `ErrInvalidCid` could just be an `errors.New("invalid cid")`.

hacdias:
@rvagg 

> Is the custom Is necessary here? The Unwrap should be enough for errors.Is to work shouldn't it?

Yes, it is necessary. You can try running the test I just added without the `.Is`. The `Unwrap` only gives you the underlying error. I want to `errors.Is(err, &ErrInvalidCid{})`. That is not possible without having the `.Is` implemented.

> It's a shame we're not at minimum of Go 1.20

Yes, I do agree that the multiple wrapping in Go 1.20 would solve much of the hassle. We've also had to implement quite a few error types in other places... Perhaps when we finally move to Go 1.20 we can change this, but then it will be a breaking change as `ErrInvalidCid` will be its own error and not just a type.

cid.go
```diff
@@ -37,10 +37,32 @@ import (
 // UnsupportedVersionString just holds an error message
 const UnsupportedVersionString = "<unsupported cid version>"
 
+// ErrInvalidCid is an error that indicates that a CID is invalid.
+type ErrInvalidCid struct {
+	Err error
+}
+
+func (e *ErrInvalidCid) Error() string {
```
rvagg, line 45: I think you should remove all the pointers here, none of them are necessary

cid.go
```diff
@@ -37,10 +37,32 @@ import (
...
+	return fmt.Sprintf("invalid cid: %s", e.Err)
+}
+
+func (e *ErrInvalidCid) Unwrap() error {
+	return e.Err
+}
+
+func (e *ErrInvalidCid) Is(err error) bool {
+	switch err.(type) {
+	case *ErrInvalidCid:
```
rvagg, line 55: if you want to be sure then check for both the pointer and non-pointer form here

cid.go
```diff
@@ -37,10 +37,32 @@ import (
...
+	default:
+		return false
+	}
+}
+
 var (
 	// ErrCidTooShort means that the cid passed to decode was not long
 	// enough to be a valid Cid
-	ErrCidTooShort = errors.New("cid too short")
+	ErrCidTooShort = &ErrInvalidCid{errors.New("cid too short")}
```
rvagg, line 65: all these should just be without the `&`

rvagg:
See https://github.com/ipfs/go-cid/commit/fc27a0c8a4e0460863e3811a3f23f8362a281ad1 which you can pull in or reimplement yourself which changes this away from pointers.

An error only needs to satisfy `interface { Error() string }`, it doesn't matter whether it's a pointer receiver or not as long as it's got that method in some form. To some degree whether you use pointers for errors or not is a matter of taste, but `errors.Is` and friends takes away arguments based on wanting to use pointers for identity (`err1 == err2`). When you make it a pointer you're going to be allocating on the heap and involving GC etc. We're not modifying anything in the struct, if the user wants to modify it then that's their call but it's done in their local version, not the version that gets bubbled all the way up from this base layer of the stack (these errors are going to be received by users _many_ frames removed from these call sites given go-cid's location in the stack!) so there's lots of pieces in between that a modification of the error might impact; so having a value rather than a pointer takes away that risk too.

I'll see if I can get some more experienced gophers to weigh in in case I'm talking out of my butt here.

kylehuntsman:
I'd like to propose a different style of wrapping all together.

It doesn't look like `ErrInvalidCid` actually holds anymore than the error it is wrapping, so there is no need for a struct.

```golang
var (
  ErrInvalidCid = errors.New("invalid cid")
  ErrCidTooShort = fmt.Errorf("%w: cid too short")
)
```
Whenever you want to return a `ErrCidTooShort`, just return `ErrCidTooShort`. You can still check it's an `ErrInvalidCid` with `errors.Is` all without having to implement any of the functions around managing an internal `Err` or wrapping/unwrapping.
```golang
err := funcThatReturnsCidTooShort()
errors.Is(err, ErrInvalidCid) {
  // do whatever
}
```

Here's a golang playground that implements what I just described.

https://go.dev/play/p/zrCM0aQthTO
```golang
package main

import (
	"errors"
	"fmt"
)

var (
	ErrInvalidCid  = errors.New("invalid cid")
	ErrCidTooShort = fmt.Errorf("%w: cid too short", ErrInvalidCid)
)

func main() {
	err1 := funcThatReturnsCidTooShort()
	if errors.Is(err1, ErrInvalidCid) {
		fmt.Printf("err1 is an ErrInvalidCid, %s\n", err1)
	}

	err2 := funcThatReturnsWrappedErrInvalidCid()
	if errors.Is(err2, ErrInvalidCid) {
		fmt.Printf("err2 is an ErrInvalidCid, %s\n", err2)
	}
}

func funcThatReturnsCidTooShort() error {
	return ErrCidTooShort
}

func funcThatReturnsWrappedErrInvalidCid() error {
	err := errors.New("some other error from somewhere else")
	return fmt.Errorf("%w: %s", ErrInvalidCid, err)
}
```

kylehuntsman:
> It's a shame we're not at minimum of Go 1.20 because fmt.Errorf("%w: %w", ErrInvalidCid, err) would be an excellent pattern to avoid the additional code here, ErrInvalidCid could just be an errors.New("invalid cid").

I just saw Rod's comment about this. This should work since [go 1.13](https://go.dev/doc/go1.13#errors) no? I feel dumb now for writing out this whole thing and then realizing you guys might have already dismissed this because of the go version. Whoops  🤷.

I see that go 1.20 supports wrapping of multiple errors, as Rod already pointed out. What is the reason that the code below doesn't work? Is it if `err` is a `ErrCustom` and then stringified, then you lose the ability to check if the returned error is also a `ErrCustom`?
```golang
return fmt.Errorf("%w, %s", ErrInvalidCid, err)
````

**Comment by rvagg** [comment]
> What is the reason that the code below doesn't work? 

All of the cases where there's currently this: `&ErrInvalidCid{err}` - wrapping an error returned from another library where you don't really get to tell the user they don't need the ability to `errors.Is` _that_ error. Without multiple wrapping you lose the ability to `errors.Is()` on whatever error is coming out of go-multibase or go-multihash or whatever hasher is involved here that may have some special error type or sentinel worth unwrapping.

Same reason `ErrInvalidCid = errors.New("invalid cid")` and `%w` isn't going to work here, because you need `%w: %w` to get both of them in the wrapping chain .. which you can't do (unless we pull in multierr but more dependencies mean more problems), so maybe we just refactor this after Go 1.21 comes out and we get to drop Go 1.19.

**Comment by kylehuntsman** [comment]
> ... so maybe we just refactor this after Go 1.21 comes out and we get to drop Go 1.19.

This might make sense. In [ipfs/go-libipfs/pull/205](https://github.com/ipfs/go-libipfs/pull/205/files#diff-a3ecfac3815e0ae44ad28cca2ea219884c8937f3de29ad4bb25b7673ae441089R141-R142), the usage is:
```golang
errors.Is(err, &cid.ErrInvalidCid{}) {
  ...
}
```
If we don't wait for multiple wrapping, then the above breaks when multiple wrapping is implemented.

We could implement a local `isErrInvalidCid()`, doing the hard work for determining if any given error is in fact a `ErrInvalidCid`. Easily backwards compatible and doesn't change the interface once multiple wraps are supported. You can just choose not to use it once we support multiple wraps. 🤔

rvagg:
Conversation here about `Is*Err()` functions (tl;dr: no): https://github.com/ipfs/go-ipld-format/pull/76

kylehuntsman:
Yeah okay, I see how this goes in circles back to implementing `Is()`, etc.

hacdias:
@rvagg what you propose is also fine, I just wanted to understand the reasoning behind it. I applied your commit.

> If we don't wait for multiple wrapping, then the above breaks when multiple wrapping is implemented.

@kylehuntsman yes, it breaks _if_ we implement multiple wrapping here _and_ release it as a patch release. If we do a 0.X release it should be fine to say there's a breaking change. I wouldn't consider that the most important piece right now.

cid_test.go
```diff
@@ -763,3 +776,15 @@ func TestLoggable(t *testing.T) {
 		t.Fatalf("did not get expected loggable form (got %v)", actual)
 	}
 }
+
+func TestErrInvalidCid(t *testing.T) {
```
masih, line 780: Can we expand this test to a table test that:
* covers all exported `cid` package functions that return value
* makes stronger assertions that include checking error message

please?

cid_test.go
```diff
@@ -763,3 +776,15 @@ func TestLoggable(t *testing.T) {
...
+	_, err := Decode("not-a-cid")
+	if err == nil {
+		t.Fatal("expected error")
+	}
+
+	is := errors.Is(err, ErrInvalidCid{})
+	if !is {
+		t.Fatal("expected error to be ErrInvalidCid")
+	}
```
masih, line 789:
```suggestion
	if !strings.HasPrefix(err.Error(), "invalid cid: ") {
		t.Fatal(`expected error message to contain "invalid cid: "`)
	}
	if !errors.Is(err, ErrInvalidCid{}) {
		t.Fatal("expected error to be ErrInvalidCid")
	}
	if !errors.Is(&ErrInvalidCid{}, ErrInvalidCid{}) {
		t.Fatal("expected error pointer to be ErrInvalidCid")
	}
```

masih (commented):
Minor comments otherwise LGTM

cid_test.go
hacdias: @masih I added some more tests. Can you please take a look? It's important to note that only `Decode`, `Cast` and `Parse` are returning `ErrInvalidCid`. Should thing such as `Prefix.Sum` also return that? It seems that the errors there are prefix-related and not CID related

masih: My vote is that the entire API should behave consistently. So if CID is invalid in any of the exposed functions then the appropriate error should be returned.

hacdias: I'll update it then

hacdias: @masih updated - afaik all exported functions that parse CIDs will return with `ErrInvalidCid`.

masih: Thank you

Add tests for the following while at it?
* `cid.CidFromBytes`
* `cid.NewCidV0`
* `cid.NewCidV1`
* `cid.PrefixFromBytes`
* `cid.ExtractEncoding`

hacdias: Why `NewCidV0` and `NewCidV1`? Those don't return errors.

The remaining ones: I added, but they're mostly duplicating existing tests (which I modified to also check for `ErrInvalidCid`).

masih: Right you are thanks @hacdias

rvagg (approved):
Thanks for jumping through the hoops @hacdias, sorry for the pile-on, we're at the bottom of the stack so it's nice to get this RIGHT!

Let's bump a 0.4.0 for this, it's technically breaking because it removes the ability to do things like `if err == mh.ErrInvalidMultihash {}` which would work with the current version. There's probably a bunch of other sentinel errors under here too, like `mbase.ErrUnsupportedEncoding`.

PR #151, opened 2023-04-01
CidFromReader should not wrap valid EOF return.

### LINKED ISSUE #152: CidFromReader wraps valid io.EOF in ErrInvalidCid
When reading from an `io.Reader` that has no data, the `io.EOF` error should not be wrapped in `ErrInvalidCid`. This is not an invalid CID, and is not the same as a partial read which is indicated by `io.ErrUnexpectedEOF`.

This is an issue because existing code that uses `CidFromReader` may check for the end of an input stream by `if err == io.EOF` instead of the preferred `if errors.Is(err, io.EOF)`, and that code will break at runtime after upgrading to go-cid v0.4.0.

gammazero:
When reading from an `io.Reader` that has no data, the `io.EOF` error should not be wrapped in `ErrInvalidCid`. This is not an invalid CID, and is not the same as a partial read which is indicated by `io.ErrUnexpectedEOF`.

This fix is needed because existing code that uses `CidFromReader` may check for the end of an input stream by `if err == io.EOF` instead of the preferred `if errors.Is(err, io.EOF)`, and that code will break at runtime after upgrading to go-cid v0.4.0.

Fixes #152

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -717,6 +717,9 @@ func (r *bufByteReader) ReadByte() (byte, error) {
 // It's recommended to supply a reader that buffers and implements io.ByteReader,
 // as CidFromReader has to do many single-byte reads to decode varints.
 // If the argument only implements io.Reader, single-byte Read calls are used instead.
+//
+// If the Reader is found to yield zero bytes, an io.EOF error is returned directly, in all
+// other error cases, an ErrInvalidCid, wrapping the original error, is returned.
 func CidFromReader(r io.Reader) (int, Cid, error) {
 	// 64 bytes is enough for any CIDv0,
 	// and it's enough for most CIDv1s in practice.
--- a/cid.go
+++ b/cid.go
@@ -728,7 +728,8 @@ func CidFromReader(r io.Reader) (int, Cid, error) {
 	vers, err := varint.ReadUvarint(br)
 	if err != nil {
 		if err == io.EOF {
-			// No data; not an invalid CID.
+			// First-byte read in ReadUvarint errors with io.EOF, so reader has no data.
+			// Subsequent reads with an EOF will return io.ErrUnexpectedEOF and be wrapped here.
 			return 0, Undef, err
 		}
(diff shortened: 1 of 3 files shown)


masih (approved):

rvagg:
This is a tough call, arguably `errors.Is(err, io.EOF)` is the right way to deal with this but your code really has limited it to the first byte, so specifically the case where you have a zero-byte reader. I think this deserves some comments which I'll suggest inline.
Would you also mind adding a case for >1 byte that fails on the varint call but doesn't return an error - `[]byte{0x80, 0x01}` should be good enough and _should_ return an `ErrInvalidCId` wrapping an `ErrUnexpectedEOF`.

cid.go
```diff
@@ -727,6 +727,10 @@ func CidFromReader(r io.Reader) (int, Cid, error) {
 	// The varint package wants a io.ByteReader, so we must wrap our io.Reader.
 	vers, err := varint.ReadUvarint(br)
 	if err != nil {
+		if err == io.EOF {
+			// No data; not an invalid CID.
```
rvagg, line 731:
```suggestion
		if err == io.EOF {
			// First-byte read in ReadUvarint errors with io.EOF, so reader has no data.
			// Subsequent reads with an EOF will return io.ErrUnexpectedEOF and be wrapped here.
```

rvagg:
Unfortunately I can't suggest inline for the godoc comment because it's too far out of the diff, but here's what I suggest for clearning up that:

```
// It's recommended to supply a reader that buffers and implements io.ByteReader,
// as CidFromReader has to do many single-byte reads to decode varints.
// If the argument only implements io.Reader, single-byte Read calls are used instead.
//
// If the Reader is found to yield zero bytes, an io.EOF error is returned directly, in all
// other error cases, an ErrInvalidCid, wrapping the original error, is returned.
func CidFromReader(r io.Reader) (int, Cid, error) {
```

rvagg (approved):
nice

PR #157, opened 2023-04-17
feat: add non-dag codec variants (json/cbor)

Stebalien:
We now CBOR, at least, in Filecoin when we want structured data that can't have links.


rvagg (approved):
ok, if we must, but I think I'd rather these consts be considered legacy and people just use https://github.com/multiformats/go-multicodec instead; it's auto generated and kept in sync with the table

aschmahmann:
@Stebalien do you still want this merged, or is go-multicodec fine?

Stebalien:
@aschmahmann are we going to deprecate the current constants?

PR #159, opened 2023-06-12
perf: add a generic CID type which allows for aliased byte slices

Jorropo:
This is backward  compatible (that why I hacked it with generics).
I should probably make generic versions of the builders too.

Results:
```
name   old time/op    new time/op    delta
PB-12     730ns ± 5%     255ns ± 2%   -65.04%  (p=0.000 n=9+10)

name   old alloc/op   new alloc/op   delta
PB-12      224B ± 0%        0B       -100.00%  (p=0.000 n=10+10)

name   old allocs/op  new allocs/op  delta
PB-12      4.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
```

### DIFF
--- a/cid.go
+++ b/cid.go
@@ -607,6 +607,11 @@ func (c CidOf[S]) Prefix() Prefix {
 	}
 }
 
+// Append appends the cid to some slice of bytes.
+func (c CidOf[S]) Append(in []byte) []byte {
+	return append(in, c.storage...)
+}
+
 // Prefix represents all the metadata of a Cid,
 // that is, the Version, the Codec, the Multihash type
 // and the Multihash length. It does not contains
--- a/cid.go
+++ b/cid.go
@@ -22,7 +22,6 @@ package cid
 import (
 	"bytes"
 	"encoding"
-	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


cid.go
```diff
@@ -170,42 +175,72 @@ var (
...
+
+// Storage defines usable backing containers for [GenericCid].
+// Using [string] allows for compile-time immutable CIDs that are comparable.
+// Using a byte slice allows for aliasable CIDs and zero allocation.
+// When a byte slice is used it MUST NOT be modified while the cid is alive.
+type Storage interface{ string | []byte }
 
-// Undef can be used to represent a nil or undefined Cid, using Cid{}
+// GenericCid is like [Cid] except it's backing array is generic.
+type GenericCid[S Storage] struct{ str S }
```
willscott, line 187: is it worth renaming `str` now that it isn't guaranteed to be a string?

cid.go
```diff
@@ -170,42 +175,72 @@ var (
 // Cid represents a self-describing content addressed
 // identifier. It is formed by a Version, a Codec (which indicates
 // a multicodec-packed content type) and a Multihash.
-type Cid struct{ str string }
+type Cid = GenericCid[string]
```
willscott, line 178: We're adding `GenericCid` to the interface - is that the right name? (it feels like it's an implementation detail leaking into the interface name - that it has a generic backing storage should be obvious from it's instantiation)

`CidFrom[string]`?
`CidOf[string]`?
`TypedCid[string]`?

willscott (commented):
* We probably should have the cid test and fuzzing also run against GenericCid
* would be good to add a test demonstrating that a cid with a backing []byte can interop with a string cid.

cid.go
Jorropo: I didn't changed `Cid` because of backward compat (this would be gonna bloody).

> We're adding `GenericCid` to the interface - is that the right name? 

Clearly not, looking for something better.
`CidOf` is ok for me.

willscott: right. i agree with the construction here for backwards compatibility 👍

MichaelMure:
Imho this is valuable and should be revisited. Code is still available at https://github.com/ipfs/go-cid/compare/468cca0d172b911fa0a8a0f59ba087682531fea6

rvagg:
@MichaelMure feel free to take it over and try again with a new PR

PR #169, opened 2024-09-24
move _rsrch/cidiface into an internal package

MichaelMure:
`_rsrch/cidiface` is an engineering experiment package to explore the right form for a CID. As such, it's not meant to be actually used, and could result in conflict if used.

To avoid accidentally importing it (or even getting suggestions to do so), this package is moved into an "internal" folder, hinting tooling that it's not meant for external consumption.


rvagg (approved):
alternatively we could just delete this folder

MichaelMure:
@rvagg FYI, I don't have the power to merge this myself.

rvagg:
yeah, sorry, just giving others a chance to pipe up before merging

